PR #3, opened 2020-07-25
First draft

spencergilbert:
Rough first draft of what a "chart per component" design could look like, based on a combination of the existing Thanos charts we identified here: https://docs.google.com/document/d/18GXxwOm9c2fDK20LV1Sekte0xvK4qXpIGwBcuw_ZYSA/edit#


charts/thanos-store/templates/ingress.yaml
```diff
@@ -0,0 +1,87 @@
+{{- if .Values.http.ingress.enabled -}}
+{{- $fullName := include "thanos-query.fullname" . -}}
```
spencergilbert, line 2: Was pointed out in slack that I didn't copy-paste this well :smile:

spencergilbert:
My biggest/only concern is if we should do a separate chart per Thanos component, or if they should all be parts of a single Thanos chart

charts/thanos-store/templates/ingress.yaml
spencergilbert: Fixed this and a few other cases in the template.

ChiefAlexander:
>My biggest/only concern is if we should do a separate chart per Thanos component

Stumbled upon this while I was looking at implementing Thanos, throwing out my own thoughts. Ideally, the end state that I would expect is that when I run a helm install all of the components that I need to run Thanos are present. Maybe the path here is to utilize helm dependences in a "main" chart that can fulfill that role, while still allowing users to utilize the underlying component charts.

Also if the path that we are going to take here is to use multiple charts I would suggest utilizing a helm library chart -> https://helm.sh/docs/topics/library_charts/

spencergilbert:
I'm going to close this in favor of PRs per component.

PR #6, opened 2020-09-18
initial thanos query-frontend chart

spencergilbert:

Initial query-frontend chart

  - fixes #

#### Checklist

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -1,5 +1,5 @@
 # consider helm install to be failed after 10 minutes
-helm-extra-args: --timeout 600
+helm-extra-args: --timeout 600s
 check-version-increment: true
 debug: true
 chart-dirs: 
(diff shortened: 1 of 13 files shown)


spencergilbert:
The changes look good to me - I'm going to get it deployed in one of our clusters

spencergilbert:
@obiesmans - all looks good and is running fine after I updated. I think the first order of business is going to be prettifying and improving the readme's but functionally I think we're in a good spot.

On the query chart I added support for initContainers and additional containers to run in the pod for people who might manually configure proxy sidecars and the like. Should we add that here as well before merging?

obiesmans:
@spencergilbert : Thanks for your review. I've added `initContainers` functionnality. I agree with you that the Readme could be more refined.

obiesmans (approved):

spencergilbert:
LGTM - are we good with merging this then?

PR #9, opened 2025-09-11
feat: add community thanos helm chart

hamdikh:

- Creates Thanos Helm Chart as Bitnami updated their support policy for their owned artifacts
- Creates Thanos Operator Helm Chart to simplify thanos-operator deployment

  - fixes #

#### Checklist


charts/thanos/values.yaml
```diff
@@ -0,0 +1,989 @@
...
+      #     hosts:
+      #       - bucketweb.local
+
+    # HorizontalPodAutoscaler for bucketweb only
+    autoscaling:
+      enabled: false
+      minReplicas: 1
+      maxReplicas: 3
+      targetCPUUtilizationPercentage: 80
+      targetMemoryUtilizationPercentage: null
```
eumel8: for all of this autoscaling feature, also if `enabled: false` I've got error on helm upgrade:

```
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
thanos:
- at '/ruler/autoscaling': missing property 'targetMemoryUtilizationPercentage'
- at '/bucket/bucketweb/autoscaling': missing property 'targetMemoryUtilizationPercentage'
- at '/queryFrontend/autoscaling': missing property 'targetMemoryUtilizationPercentage'
- at '/receive/autoscaling': missing property 'targetMemoryUtilizationPercentage'
- at '/query/autoscaling': missing property 'targetMemoryUtilizationPercentage'
- at '/storegateway/autoscaling': missing property 'targetMemoryUtilizationPercentage'
```

maybe `null` is not allowed? for simplification it should the default like the CPU percentage

charts/thanos/Chart.yaml
```diff
@@ -0,0 +1,31 @@
...
+icon: https://raw.githubusercontent.com/thanos-io/thanos/master/docs/img/Thanos-logo_fullmedium.png
+deprecated: false
+kubeVersion: ">= 1.30.0-0"
+maintainers:
+  - name: hamdikh
+    email: hamdi.khelil@kapheira.cloud
+dependencies:
+  - name: kube-prometheus-stack
+    alias: kube-prometheus-stack
+    version: 77.6.0
```
eumel8: should be updated to 79.7.1, there is an issue with killing Pods, possibly due label mismatch

charts/thanos/values.yaml
```diff
@@ -0,0 +1,989 @@
...
+  priorityClassName: ""
+
+  annotations: {}
+  labels: {}
+
+  extraVolumes: []
+  extraVolumeMounts: []
+
+  serviceMonitor:
+    enabled: true
```
eumel8: sometimes the user decided not to use Prometheus Operator and set
```
kube-prometheus-stack:
  crds:
    enabled: false
```
for this reason I think serviceMonitor should be always disabled by default.

charts/thanos/Chart.yaml
hamdikh: this will be managed automatically by dependabot or renovate, I'll add the latest version available as of now => v80.2.0

Issue #44, opened 2026-04-13
bug: multiple values fields are not wired up to the components' deployment/statefulset manifests


marcdavoli:
## Environment

| Field              | Value |
|--------------------|-------|
| Chart version      |    0.3.1   |
| Thanos version     |   0.41.0   |
| Helm version       |   4.1.1    |
| Kubernetes version |   1.35.2    |
| Cloud / Platform   |     Amazon EC2  |

## Values used

```yaml
global:
  priorityClassName: critical
compactor:
  priorityClassName: high
```

## What happened

The `priorityClassName` field **does not** get added to the spec of the StatefulSets/Deployments deployed by the Helm Chart.

## What you expected to happen

The `priorityClassName` field should get added to the spec of the StatefulSets/Deployments deployed by the Helm Chart.

## How to reproduce it

1. Deploy any component with targeted Helm Chart version
2. Add the values `global.priorityClassName` and/or `<component>.priorityClassName`
3. Observe that the change is not applied to the component(s)

## Relevant logs

No logs produced, the change is a no-op.

## Anything else

I've observed that there are many fields like this that are not wired up properly to the Deployment/StatefulSet manifests:
- priorityClassName
- nodeSelector
- tolerations
- affinity
- topologySpreadConstraints

PR #45, opened 2026-04-14
Additions and fixes

jim-barber-he:

The chart as it is now seems to be incomplete since many of the values defined are not used, so I've changed that.
I found a bug where the query command does not accept the --store command line parameter.
I added a missing Service for Thanos Ruler.
I added optional NetworkPolicy resources.
I made it so appropriate pods roll if the object store configuration is modified.
Other minor changes mentioned in my main write up below:

Commit message follows:

A number of the global.* and $component.* variables defined in values.yaml were not actually used in the templates. Update the templates so that they are used.
The following variables now have an effect where they didn't before:
- *.affinity
- *.extraContainers
- *.extraEnv
- *.extraEnvFrom
- *.extraInitContainers
- *.nodeSelector
- *.priorityClassName
- *.tolerations
- *.topologySpreadConstraints
- bucket.bucketweb.replicaCount
- global.pdb
- $component.annotations
- $component.labels
- $component.service.annotations
- $component.service.labels

Add the ability to configure the dnsConfig field of the pods.

Add optional NetworkPolicy resources driven by a new global.networkPolicies boolean value that defaults to false.

Add a Service for Thanos ruler.

Fix that the query pod does not understand the `--store` command line option. Replace it with `--endpoint` instead.

Have the *.pdb.maxUnavailable and *.pdb.minAvailable variables accept both integer and string values.

Use port names instead of numbers in the Ingress resources and the probes. Note that the thanos.httpProbes template could be simplified to no longer receive a port parameter and just always use the http port name, however I didn't change that in case the flexibility is still required at some point.

Add a checksum/objstore-configuration annotation to resources that mount the object store configuration so that pods roll if it is changed.

Replace indent with nindent in templates and add leading indents changing the initial {{ to {{- so that the templates are indented nicely.

- fixes #

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.3.1
+version: 0.3.2
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.2.0](https://img.shields.io/badge/Version-0.2.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.39.2](https://img.shields.io/badge/AppVersion-v0.39.2-informational?style=flat-square)
+![Version: 0.3.2](https://img.shields.io/badge/Version-0.3.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
@@ -45,8 +45,8 @@ Kubernetes: `>= 1.30.0-0`
 
 | Repository | Name | Version |
 |------------|------|---------|
-| https://charts.rustfs.com/ | rustfs | 0.0.90 |
-| https://prometheus-community.github.io/helm-charts | kube-prometheus-stack(kube-prometheus-stack) | 80.2.0 |
+| https://charts.rustfs.com/ | rustfs | 0.0.91 |
+| https://prometheus-community.github.io/helm-charts | kube-prometheus-stack(kube-prometheus-stack) | 83.0.2 |
 
 ## Component Overview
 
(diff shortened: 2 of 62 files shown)


jim-barber-he:
@hamdikh 

I'm not sure if you're happy with this being one large PR or if you require me to break it up into separate PRs.
I'm hoping you'll accept this PR since I've found no decent alternative Helm chart for Thanos after moving away from the Bitnami Helm chart.
I noticed this chart has had recent activity again, so thought I'd get it working on my environment and contribute back.

jim-barber-he:
Looking at the existing issues this should address this one: https://github.com/thanos-community/helm-charts/issues/44

hamdikh:
Hey @jim-barber-he  

First of all thank you for the awesome work and contributions, really appreciate it!

One thing that would help a lot with the review process: could you split this PR into multiple ones? Ideally, fixes should be in their own dedicated PR(s), and additions should be addressed either one issue at a time or grouped by feature type. This makes it much easier to review and reason about each change independently.

Thanks again! 🔥🔥🔥

jim-barber-he:
@hamdikh Yeah okay. I can do that. I needed another feature which i just pushed that could be done separately as well.

jim-barber-he:
@hamdikh A question on splitting up the PRs. 
Is one PR covering the handling of implementing the missing values within the templates okay?
Or will you want one for affinity; one for extraContainers (and possibly initContainers lumped with that); one for extraEnv and extraEnvFrom; and so on?

hamdikh:
> @hamdikh A question on splitting up the PRs. Is one PR covering the handling of implementing the missing values within the templates okay? Or will you want one for affinity; one for extraContainers (and possibly initContainers lumped with that); one for extraEnv and extraEnvFrom; and so on?

One PR will do fine, if you could as well add helm unittests against the additions it'll be awsome

hamdikh:
> @hamdikh Yeah okay. I can do that. I needed another feature which i just pushed that could be done separately as well.

@jim-barber-he I started reading the PR content, for TLS we might want to add it across all components, I suggest embeding the config directly into the existing components for that (tests should be written as well, don't hesitate to ping me if you want a helping hand with anything)

Thanks again for the contributions !! 🥇

jim-barber-he:
> for TLS we might want to add it across all components, I suggest embeding the config directly into the existing components for that 

@hamdikh Are you talking about the new `query-tls` Deployment I added?
The reason I did that is I have need to talk to remote clusters that are running the kube-prometheus-stack with the Thanos gateway sidecar containers.
The access to these are via a TLS load balancer.

For the existing chart, if I try to add the remove stores I end up with these `--endpoint` parameters being passed to the `thanos query` command:
```
--endpoint=dnssrv+_grpc._tcp.thanos-storegateway.thanos.svc.cluster.local
--endpoint=thanos-gateway.test1.apps.he0.io:443
--endpoint=thanos-gateway.test2.apps.he0.io:443
```

BUT the `thanos query` command can only deal with all the endpoints being TLS enabled, or all of them not being TLS enabled.
The above is a mix of two.

My solution was to run another query Deployment (that I called `query-tls`) to talk to the TLS endpoints and add that as an endpoint to the non-TLS endpoint.
So the above becomes:
```
--endpoint=dnssrv+_grpc._tcp.thanos-storegateway.thanos.svc.cluster.local
--endpoint=dnssrv+_grpc._tcp.thanos-query-tls.thanos.svc.cluster.local
```

And then the query-tls Deployment has:
```
--endpoint=thanos-gateway.test1.apps.he0.io:443
--endpoint=thanos-gateway.test2.apps.he0.io:443
```

*Updated* Accidentally pressed Enter while still typing this out.

PR #55, opened 2026-04-15
feat(dnsconfig): Add support to set pod dnsConfig

jim-barber-he:
#### What this PR does / why we need it:

Add a new dnsConfig variable to the global variables and the variables for each component in the thanos Helm chart.
This will set the pod.spec.dnsConfig on the pods.

Add helm unit tests for the new variable.

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #

#### Checklist

[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]

- [x] [DCO](https://developercertificate.org) signed
- [x] Chart Version bumped (if the pr is an update to an existing chart)
- [x] Variables are documented in the README.md

PR #59, opened 2026-04-16
feat(networkpolicy): Add optional NetworkPolicy resources

jim-barber-he:

Add network policies for each of the components that will be created if the new global.networkPolicies value is set to true.

Add unittests for handling when the new network policies are created or not.

Use port names instead of numbers for the probes and update the validation schema to accept both integers and strings for the ports.

Use port names instead of numbers for the Ingress resources.

- fixes #

#### Checklist

### DIFF
--- a/charts/thanos/templates/query-frontend/ingress.yaml
+++ b/charts/thanos/templates/query-frontend/ingress.yaml
@@ -24,7 +24,7 @@ spec:
               service:
                 name: {{ include "thanos.compName" (list $ "query-frontend") }}
                 port:
-                  number: {{ $.Values.queryFrontend.service.port }}
+                  name: http
         {{- end }}
   {{- end }}
   tls:
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.4.5
+version: 0.5.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.4.4](https://img.shields.io/badge/Version-0.4.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.5.0](https://img.shields.io/badge/Version-0.5.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
@@ -472,7 +472,7 @@ The table below documents all available values. Top-level keys group settings by
 | bucket.bucketweb.probes.liveness.initialDelaySeconds | int | `30` | Seconds to wait before starting the liveness probe. |
 | bucket.bucketweb.probes.liveness.path | string | `"/-/healthy"` | HTTP path checked by the liveness probe. |
 | bucket.bucketweb.probes.liveness.periodSeconds | int | `10` | How often (seconds) to run the liveness probe. |
-| bucket.bucketweb.probes.liveness.port | int | `10902` | Port checked by the liveness probe. |
+| bucket.bucketweb.probes.liveness.port | int or string | `"http"` | Port checked by the liveness probe. |
 | bucket.bucketweb.probes.liveness.successThreshold | int | `1` | Consecutive successes before the container is considered live. |
 | bucket.bucketweb.probes.liveness.timeoutSeconds | int | `5` | Seconds after which the probe times out. |
 | bucket.bucketweb.probes.readiness.enabled | bool | `true` | Enable the readiness probe for Bucketweb. |
@@ -480,7 +480,7 @@ The table below documents all available values. Top-level keys group settings by
 | bucket.bucketweb.probes.readiness.initialDelaySeconds | int | `5` | Seconds to wait before starting the readiness probe. |
 | bucket.bucketweb.probes.readiness.path | string | `"/-/ready"` | HTTP path checked by the readiness probe. |
 | bucket.bucketweb.probes.readiness.periodSeconds | int | `10` | How often (seconds) to run the readiness probe. |
-| bucket.bucketweb.probes.readiness.port | int | `10902` | Port checked by the readiness probe. |
+| bucket.bucketweb.probes.readiness.port | int or string | `"http"` | Port checked by the readiness probe. |
 | bucket.bucketweb.probes.readiness.successThreshold | int | `1` | Consecutive successes before the pod is marked ready. |
 | bucket.bucketweb.probes.readiness.timeoutSeconds | int | `5` | Seconds after which the probe times out. |
 | bucket.bucketweb.probes.startup.enabled | bool | `true` | Enable the startup probe for Bucketweb. |
@@ -488,7 +488,7 @@ The table below documents all available values. Top-level keys group settings by
 | bucket.bucketweb.probes.startup.initialDelaySeconds | int | `0` | Seconds to wait before starting the startup probe. |
 | bucket.bucketweb.probes.startup.path | string | `"/-/ready"` | HTTP path checked by the startup probe. |
 | bucket.bucketweb.probes.startup.periodSeconds | int | `5` | How often (seconds) to run the startup probe. |
-| bucket.bucketweb.probes.startup.port | int | `10902` | Port checked by the startup probe. |
+| bucket.bucketweb.probes.startup.port | int or string | `"http"` | Port checked by the startup probe. |
 | bucket.bucketweb.probes.startup.successThreshold | int | `1` | Consecutive successes required before the startup probe is considered passed. |
 | bucket.bucketweb.probes.startup.timeoutSeconds | int | `5` | Seconds after which the probe times out. |
 | bucket.bucketweb.replicaCount | int | `1` | Number of Bucketweb pod replicas. |
@@ -554,7 +554,7 @@ The table below documents all available values. Top-level keys group settings by
 | compactor.probes.liveness.initialDelaySeconds | int | `30` | Seconds to wait before starting the Compactor liveness probe. |
 | compactor.probes.liveness.path | string | `"/-/healthy"` | HTTP path checked by the Compactor liveness probe. |
 | compactor.probes.liveness.periodSeconds | int | `10` | How often (seconds) to run the Compactor liveness probe. |
-| compactor.probes.liveness.port | int | `10902` | Port checked by the Compactor liveness probe. |
+| compactor.probes.liveness.port | int or string | `"http"` | Port checked by the Compactor liveness probe. |
 | compactor.probes.liveness.successThreshold | int | `1` | Consecutive successes before the Compactor container is considered live. |
 | compactor.probes.liveness.timeoutSeconds | int | `5` | Seconds after which the Compactor liveness probe times out. |
 | compactor.probes.readiness.enabled | bool | `true` | Enable the readiness probe for the Compactor. |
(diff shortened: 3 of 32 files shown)


jim-barber-he:
@hamdikh The next feature to be added after the dnsConfig PR https://github.com/thanos-community/helm-charts/pull/55

charts/thanos/templates/query-frontend/ingress.yaml
```diff
@@ -24,7 +24,7 @@ spec:
               service:
                 name: {{ include "thanos.compName" (list $ "query-frontend") }}
                 port:
-                  number: {{ $.Values.queryFrontend.service.port }}
+                  name: http
```
hamdikh, line 27: Why are we switching back from using port number to port name ?

hamdikh (changes requested):
Thanks for the NetworkPolicy addition, the feature itself is welcome and the structure looks good. However, this PR introduces breaking changes that are unrelated to NetworkPolicies and need to be addressed before merging.

Across all components (bucket, compactor, query, query-frontend, receive, ruler, storegateway), the PR replaces dynamic port references with hardcoded named ports in three different places:

1. Ingress templates: `port.number: {{ .Values.*.service.port }}` has been replaced with `port.name: http`. This breaks ingress routing for any user who configures a non-default port, since the named port `http` may not exist on their Service.

2. Liveness/readiness probes: `thanos.httpProbes` now receives the string `"http"` instead of the numeric port from values. This silently breaks health checks when the container port is not explicitly named `http`.

3. NetworkPolicy port references: ingress rules use hardcoded `port: http` and `port: grpc` instead of referencing configured values.

Additionally, the `receive` component previously had a safe default fallback (`default 10902 .Values.receive.service.httpPort`) that has been removed entirely.

These changes are undocumented, not mentioned in the PR description, and will silently break deployments for users with custom port configurations. The NetworkPolicy templates themselves are fine, but the port reference changes need to be reverted.

Requested changes:
- Revert all `port.name: http` in ingress templates back to `port.number: {{ .Values.*.service.port }}`
- Revert `thanos.httpProbes` calls back to passing the numeric port from values
- Restore the `default 10902` fallback for the receive component
- If named ports are intentional for NetworkPolicies, document the assumption and ensure Service templates consistently define port names

charts/thanos/templates/query-frontend/ingress.yaml
jim-barber-he: I have read in the past that it's best practice to use names over numbers, and to me it makes sense because:

- When a port number changes you only have to change it in one place (the pod definition).
  There's no need to update the probes, services, and ingresses as well.

- Readability.
  When looking at the manifests you don't have to memorise that port 10901 is the GRPC port while port 10902 is the HTTP port for some containers while port 9090 is the HTTP port on other containers...

- It simplifies the Helm chart since you no longer have to reference `{{ $.Values.COMPONENT.service.port }}` all over the place.
  There's less chance of making a mistake going forward.

The official document talking about named ports have a quote about being able to then change port numbers without breaking clients:
https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
> This works even if there is a mixture of Pods in the Service using a single configured name, with the same network protocol available via different port numbers. This offers a lot of flexibility for deploying and evolving your Services. For example, you can change the port numbers that Pods expose in the next version of your backend software, without breaking clients.

But in the end it's just a preference of mine and you can reject it if you want, but I think it has benefits.

jim-barber-he: I'll try to address your concerns you raised in the first comment.

> Across all components (bucket, compactor, query, query-frontend, receive, ruler, storegateway), the PR replaces dynamic port references with hardcoded named ports in three different places:
> 
>   1.  Ingress templates: port.number: {{ .Values.*.service.port }} has been replaced with port.name: http. This breaks ingress routing for any user who configures a non-default port, since the named port http may not exist on their Service.

The Ingress change refers to the `http` port name in the Service which is hardcoded in the `name` field and isn't configurable by the user.
The configurable `port` is the actual port number that is customised by the user.
You'll note that the Service passes traffic through to the `targetPort` which also happens to be a port name (this wasn't my change) and in the containers themselves the port named `http` that targetPort is pointing at has the port number as configured by the user (or the default of the chart if the user hasn't set it).
```
=== charts/thanos/templates/bucket/service.yaml
  ports:
    - name: http
      port: {{ .Values.bucket.bucketweb.service.port }}
      targetPort: http

=== charts/thanos/templates/compactor/service.yaml
  ports:
    - name: http
      port: {{ .Values.compactor.service.port }}
      targetPort: http

=== charts/thanos/templates/query-frontend/service.yaml
  ports:
    - name: http
      port: {{ .Values.queryFrontend.service.port }}
      targetPort: http

=== charts/thanos/templates/query/service.yaml
  ports:
    - name: http
      port: {{ .Values.query.service.httpPort }}
      targetPort: http
    - name: grpc
      port: {{ .Values.query.service.grpcPort }}
      targetPort: grpc

=== charts/thanos/templates/receive/service.yaml
  ports:
    - name: http
      port: {{ .Values.receive.service.httpPort }}
      targetPort: http
    - name: remote-write
      port: {{ .Values.receive.service.remoteWritePort }}
      targetPort: remote-write
  ports:
    - name: grpc
      port: {{ .Values.receive.service.grpcPort }}
      targetPort: grpc

=== charts/thanos/templates/storegateway/service.yaml
  ports:
    - name: http
      port: {{ .Values.storegateway.service.httpPort }}
      targetPort: http
    - name: grpc
      port: {{ .Values.storegateway.service.grpcPort }}
      targetPort: grpc
```
All of the `.Values.$component.service.*` values used by the `port` fields are port numbers in the values.yaml file:
```
$ for i in bucket.bucketweb.service.port compactor.service.port queryFrontend.service.port query.service.httpPort query.service.grpcPort receive.service.httpPort receive.service.remoteWritePort receive.service.grpcPort storegateway.service.httpPort storegateway.service.grpcPort; do yq .$i charts/thanos/values.yaml; done
10902
10902
9090
9090
10901
10902
10908
10901
10902
10901
```

>   2.  Liveness/readiness probes: thanos.httpProbes now receives the string "http" instead of the numeric port from values. This silently breaks health checks when the container port is not explicitly named http.

The `http` name is hardcoded in the container definitions and doesn't change and isn't configurable by the user.
e.g. within `charts/thanos/templates/query/deployment.yaml`
```
          ports:
            - name: http
              containerPort: {{ .Values.query.service.httpPort }}
            - name: grpc
              containerPort: {{ .Values.query.service.grpcPort }}
```
So the probe hits the named port `http` and goes to whatever `containerPort` is set to (and is whatever the user set that to; or the default if not changed).

>   3. NetworkPolicy port references: ingress rules use hardcoded port: http and port: grpc instead of referencing configured values.

This is the same concept as above as well. It targets the hardcoded port names in the Services that point at the configured port number in the Service's `port` value.

> Additionally, the receive component previously had a safe default fallback (default 10902 .Values.receive.service.httpPort) that has been removed entirely.

I guess you're referring to this:
```
          {{- include "thanos.httpProbes" (dict "root" . "key" "receive" "port" (default 10902 .Values.receive.service.httpPort)) | nindent 10 }}
```
changing to:
```
          {{- include "thanos.httpProbes" (dict "root" . "key" "receive" "port" "http") | nindent 10 }}
```
Having the fallback is unnecessary because the port definition in the statefulset (`charts/thanos/templates/receive/statefulset.yaml`) uses the default fallback method.
The `http` named port that we're pointing at uses: `containerPort: {{ default 10902 .Values.receive.service.httpPort }}`
```
          ports:
            - name: grpc
              containerPort: {{ .Values.receive.service.grpcPort }}
            - name: http
              containerPort: {{ default 10902 .Values.receive.service.httpPort }}
            - name: remote-write
              containerPort: {{ .Values.receive.service.remoteWritePort }}
```
So that's another example where using port names instead of numbers has simplified things. You only need to worry about that default fallback in one place on the chart.

> These changes are undocumented, not mentioned in the PR description, and will silently break deployments for users with custom port configurations. The NetworkPolicy templates themselves are fine, but the port reference changes need to be reverted.

I mentioned the changes like so in the PR description:
```
Use port names instead of numbers for the probes and update the validation schema to accept both integers and strings for the ports.

Use port names instead of numbers for the Ingress resources.
```
But I'm happy to try and expand on them if you need me to?

> Requested changes:
> 
>     Revert all port.name: http in ingress templates back to port.number: {{ .Values.*.service.port }}
>     Revert thanos.httpProbes calls back to passing the numeric port from values
>     Restore the default 10902 fallback for the receive component
>     If named ports are intentional for NetworkPolicies, document the assumption and ensure Service templates consistently define port names

So I believe these changes haven't broken anything and have simplified things...
I have these changes deployed for real locally and it's all working fine. 

Unless I'm missing something, I do not believe these to be breaking changes.
Let me know if I am wrong after reading the above and pointing out any misunderstandings that I might have.

hamdikh: Those are all valid points and I agree that named ports are the better long-term approach for the reasons you described. The readability and single-point-of-change arguments are solid, and this is indeed the Kubernetes-recommended pattern.

That said, my concern is not with named ports as a concept, but with removing the ability for users to choose. By hardcoding `name: http` everywhere, we are implicitly forcing all users to name their ports consistently, which may not match existing deployments. Someone who has been running this chart with custom port numbers and no named ports will silently break on upgrade with no clear error message pointing them to the cause.

A middle ground would be to introduce a `portName` value per component (e.g. `compactor.service.portName`) that defaults to `http`, while keeping the numeric fallback when no name is set. That way users who want named ports get them by default, and users with existing non-standard configurations are not broken.

Something along these lines:

```yaml
port:
  {{- if .Values.compactor.service.portName }}
  name: {{ .Values.compactor.service.portName }}
  {{- else }}
  number: {{ .Values.compactor.service.port }}
  {{- end }}
```

This keeps the spirit of your change while giving users the flexibility to opt out if needed. Happy to discuss further or help iterate on the approach.

jim-barber-he: > That said, my concern is not with named ports as a concept, **but with removing the ability for users to choose**

The port names were already hardcoded in the chart as `http` though and never were configurable? So I haven't removed the user's ability to choose since they never had the choice?
I didn't change the `name` field of the port in the Services nor in the Deployments & StatefulSets.
If I check out commit `8bb077cc8275ee013ce8c849f5f1d493c64153c8` from before any of my PRs came along we have:
```
$ sed -n '/^[[:space:]]*ports:/,/^[[:space:]]*selector:/ { /^[[:space:]]*selector:/d; p }' charts/thanos/templates/*/service.yaml
  ports:
    - name: http
      port: {{ .Values.bucket.bucketweb.service.port }}
      targetPort: http
  ports:
    - name: http
      port: {{ .Values.compactor.service.port }}
      targetPort: http
  ports:
    - name: http
      port: {{ .Values.queryFrontend.service.port }}
      targetPort: http
  ports:
    - name: http
      port: {{ .Values.query.service.httpPort }}
      targetPort: http
    - name: grpc
      port: {{ .Values.query.service.grpcPort }}
      targetPort: grpc
  ports:
    - name: http
      port: {{ .Values.receive.service.httpPort }}
      targetPort: http
    - name: remote-write
      port: {{ .Values.receive.service.remoteWritePort }}
      targetPort: remote-write
  ports:
    - name: grpc
      port: {{ .Values.receive.service.grpcPort }}
      targetPort: grpc
  ports:
    - name: http
      port: {{ .Values.storegateway.service.httpPort }}
      targetPort: http
    - name: grpc
      port: {{ .Values.storegateway.service.grpcPort }}
      targetPort: grpc
```
And
```
$ for i in charts/thanos/templates/*/deployment.yaml; do echo "=== $i"; awk '/^[[:space:]]*ports:/ {b=match($0,/[^ ]/)-1;print;inside=1;next} inside {c=match($0,/[^ ]/)-1; if(c>b) print; else exit}' "$i" $i; echo; done
=== charts/thanos/templates/bucket/deployment.yaml
          ports:
            - name: http
              containerPort: {{ .Values.bucket.bucketweb.service.port }}

=== charts/thanos/templates/query/deployment.yaml
          ports:
            - name: http
              containerPort: {{ .Values.query.service.httpPort }}
            - name: grpc
              containerPort: {{ .Values.query.service.grpcPort }}

=== charts/thanos/templates/query-frontend/deployment.yaml
          ports:
            - name: http
              containerPort: {{ .Values.queryFrontend.service.port }}
```
And
```
$ for i in charts/thanos/templates/*/statefulset.yaml; do echo "=== $i"; awk '/^[[:space:]]*ports:/ {b=match($0,/[^ ]/)-1;print;inside=1;next} inside {c=match($0,/[^ ]/)-1; if(c>b) print; else exit}' "$i" $i; echo; done
=== charts/thanos/templates/compactor/statefulset.yaml
          ports:
            - name: http
              containerPort: {{ .Values.compactor.service.port }}

=== charts/thanos/templates/receive/statefulset.yaml
          ports:
            - name: grpc
              containerPort: {{ .Values.receive.service.grpcPort }}
            - name: http
              containerPort: {{ default 10902 .Values.receive.service.httpPort }}
            - name: remote-write
              containerPort: {{ .Values.receive.service.remoteWritePort }}

=== charts/thanos/templates/ruler/statefulset.yaml
          ports:
            - name: http
              containerPort: {{ .Values.ruler.service.httpPort }}

=== charts/thanos/templates/storegateway/statefulset.yaml
          ports:
            - name: grpc
              containerPort: {{ .Values.storegateway.service.grpcPort }}
            - name: http
              containerPort: {{ .Values.storegateway.service.httpPort }}
```
The names have been hardcoded the whole time, and letting  the users configure them provides no benefit?

hamdikh: valid point my bad, let me take a deeper look into the PR, i'll keep you posted 

Thank you for the great effort @jim-barber-he 🚀

jim-barber-he: I've enjoyed the discussion and it made me at least revisit a few things to make sure my understanding was correct. :laughing:

jim-barber-he:
I'll fix the conflicts after https://github.com/thanos-community/helm-charts/pull/55 is merged, since every merged PR causes a conflict with Chart.yaml and README.md when bumping the chart version.

jim-barber-he:
I debated modifying the `thanos.httpProbes` template to remove the `.port` parameter and just hardcode `http` there and stop passing it in everywhere it is called.
But decided to keep it flexible in case you ever wanted to probe some other port one day...

jim-barber-he:
@hamdikh Is there any more action you need from me at this point?

hamdikh:
@jim-barber-he  looks awsome, one more rebase i'll validate it just afterwards 🔥🔥🔥

jim-barber-he:
Rebased.

hamdikh (approved):
LGTM 👍🏽

PR #69, opened 2026-04-20
fix(ruler): incorrectly formatted ConfigMap for Prometheus Rule auto-import

### LINKED ISSUE #68: bug: Ruler produces an incorrectly formatted ConfigMap for auto-importing PrometheusRules
## Environment

| Chart version      |    0.3.1   |
| Thanos version     |    0.41.0   |
| Helm version       |    4.1.1   |
| Kubernetes version |    1.35.2   |
| Cloud / Platform   |    Amazon EC2   |

## Values used

```yaml
  ruler:
    enabled: true
    # Enabled by default via default values:
    # autoImportPrometheusRules:
    #   enabled: true
```

## What happened

I encountered 3 separate problems with the same ConfigMap/script.

### First problem

The `prometheus-rule-importer` container in the Thanos Ruler pod throws this error:
```
Using label selector: map[:]
Syncing PrometheusRules
Error from server (BadRequest): Unable to find "monitoring.coreos.com/v1, Resource=prometheusrules" that match label selector "map[]", field selector "": unable to parse requirement: <nil>: Invalid value: "map[]": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
```

The default value for `ruler.autoImportPrometheusRules.labelSelector: {}` renders to `map[]` rather than the expected `""`.
That makes the script fail at the `kubectl get prometheusrules.monitoring.coreos.com` step, later on.

### Second problem

The `prometheus-rule-importer` container in the Thanos Ruler pod throws this error:
```
No label selector provided, importing ALL PrometheusRules
Syncing PrometheusRules
Processing thanos/cluster-test-probes
/scripts/import.sh: line 27: yq: not found
```

We should not rely on `yq` being part of the `alpine/kubectl` image.
We can accomplish the same thing with `kubectl get ... -o jsonpath` and eliminate that dependency.

### Third problem

When testing changes the script to workaround the 2 previous issues above, I was able to get the script to finish successfully, but the rules still weren't being reloaded in Thanos Ruler. We need to add a step to the script that calls `curl -sf -X POST http://localhost:10902/-/reload`.

The Ruler pod should start normally and import all of the existing `PrometheusRules` in the cluster.

## How to reproduce it

1. Enable the Ruler.
2. Add a `PrometheusRule` custom resource to the cluster.
3. Watch the new Ruler pod logs (`prometheus-rule-importer` container) and observe its logs.

marcdavoli:

Fixes the formatting/syntax of the produced ConfigMap/script for PrometheusRules auto-reloading, with the suffix `-rule-importer` (e.g.: `thanos-RELEASE-NAME-ruler-rule-importer`). This should allow Thanos Ruler pods to import PrometheusRules properly, when no label selector is specified (the default behaviour).

- fixes #68

#### Checklist

### DIFF
--- a/charts/thanos/templates/ruler/configmap-auto-import-prometheus-rules.yaml
+++ b/charts/thanos/templates/ruler/configmap-auto-import-prometheus-rules.yaml
@@ -7,15 +7,17 @@ data:
   import.sh: |
     set -e
 
-    SELECTOR="{{ .Values.ruler.autoImportPrometheusRules.labelSelector }}"
+    SELECTOR="{{ include "thanos.ruler.autoImportPrometheusRules.labelSelector" . }}"
 
-    if [ -z "$SELECTOR" ] || [ "$SELECTOR" = "{}" ]; then
+    if [ -z "$SELECTOR" ]; then
       echo "No label selector provided, importing ALL PrometheusRules"
       SELECTOR_ARG=""
     else
       echo "Using label selector: $SELECTOR"
       SELECTOR_ARG="-l $SELECTOR"
     fi
+
+    PREV_SUM=""
 
     while true; do
       echo "Syncing PrometheusRules"
@@ -30,9 +32,19 @@ data:
 
         echo "Processing $ns/$name"
 
-        kubectl get prometheusrule "$name" -n "$ns" -o yaml | \
-          yq -y '{groups: .spec.groups}' > "/generated/${ns}-${name}.yaml"
+        kubectl get prometheusrule "$name" -n "$ns" -o jsonpath='{.spec}' \
+          > "/generated/${ns}-${name}.yaml"
       done
+
+      CURRENT_SUM=$(find /generated -type f -name '*.yaml' -exec cat {} + 2>/dev/null | md5sum | cut -d' ' -f1)
+      if [ "$CURRENT_SUM" != "$PREV_SUM" ]; then
+        echo "Rule files changed, triggering Thanos Ruler reload"
+        if curl -sf -X POST http://localhost:10902/-/reload; then
+          PREV_SUM="$CURRENT_SUM"
+        else
+          echo "Failed to reload Thanos Ruler, will retry next cycle"
+        fi
+      fi
 
       sleep 60
     done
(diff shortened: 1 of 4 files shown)


charts/thanos/templates/ruler/configmap-auto-import-prometheus-rules.yaml
```diff
@@ -30,10 +32,20 @@ data:
...
 
+      CURRENT_SUM=$(find /generated -type f -name '*.yaml' -exec cat {} + 2>/dev/null | md5sum | cut -d' ' -f1)
+      if [ "$CURRENT_SUM" != "$PREV_SUM" ]; then
+        echo "Rule files changed, triggering Thanos Ruler reload"
+        if curl -sf -X POST http://localhost:10902/-/reload; then
+          PREV_SUM="$CURRENT_SUM"
+        else
+          echo "Failed to reload Thanos Ruler, will retry next cycle"
+        fi
+      fi
```
marcdavoli, line 47: Adds the "reload" step to the script loop, so that Thanos Ruler can evaluate new rules without being restarted.

charts/thanos/templates/ruler/configmap-auto-import-prometheus-rules.yaml
```diff
@@ -30,10 +32,20 @@ data:
 
         echo "Processing $ns/$name"
 
-        kubectl get prometheusrule "$name" -n "$ns" -o yaml | \
-          yq -y '{groups: .spec.groups}' > "/generated/${ns}-${name}.yaml"
+        kubectl get prometheusrule "$name" -n "$ns" -o jsonpath='{.spec}' \
+          > "/generated/${ns}-${name}.yaml"
```
marcdavoli, line 36: Eliminates the dependency on `yq`, which doesn't exist in the `alpine/kubectl` Docker image.

hamdikh (approved):
LGTM 👍🏽

PR #70, opened 2026-04-21
feat(tls-endpoints): Query TLS enabled endpoints

jim-barber-he:

Provide a way to query TLS enabled endpoints.

The `thanos query` command cannot accept a mix of TLS and non-TLS enabled endpoints.

The Query component already handles the local non-TLS endpoints by talking directly to the Kubernetes services created by the chart, and optionally any other non-TLS endpoints added by users to the `query.stores[]` value.

Create a new optional query-tls component that has the job of querying TLS enabled endpoints. This is so Thanos components in other remote clusters that are only exposed by TLS enabled Ingress can be queried.

If the query-tls component is enabled, the query component will automatically add it to its list of endpoints.

- fixes #

I'm not sure if you're happy with the name `query-tls` or if I should have called it something like `query-tls-endpoints` instead to make it clearer what this query deployment is for?

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.5.0
+version: 0.6.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.5.0](https://img.shields.io/badge/Version-0.5.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.6.0](https://img.shields.io/badge/Version-0.6.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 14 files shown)


jim-barber-he:
@hamdikh Here is a feature I require to be able to query TLS endpoints.

This seemed like the cleanest way to do things to me...
Other solutions I've seen are people deploying something like an Envoy sidecar that acts as a TLS proxy and have services expose those and configure the query component to just use TLS for everything.
It results in less query deployments, but then you have to bring in something external for the sidecar...

hamdikh (changes requested):
Thanks for the work here @jim-barber-he, but I'd push back on merging this as-is.

The root cause is an upstream thanos query CLI limitation (uniform client TLS across all --endpoint values). Working around it by forking the component into a second Deployment doubles the chart surface, adds a gRPC hop to every query, and couples two concerns (endpoint federation and transport security) that should stay separate.

I'd prefer one of:

A uniform query.tls block on the existing query component (--grpc-client-tls-secure + ca/cert/key), making the chart a TLS-everywhere or TLS-nowhere deployment. ~20 lines vs ~500.
Documenting a sidecar pattern (envoy) for users who genuinely need to bridge plaintext and TLS endpoints, rather than shipping a second querier in-tree.
Pushing the fix upstream to thanos query so per-endpoint TLS config becomes a first-class CLI feature. That removes the need for any of this entirely.
If we do merge something chart-side, option 1 is the one I'd support.
The mixed-endpoint case is real but narrow and IMO belongs in the user's own overlay, not in the canonical chart.

jim-barber-he:
Ok.

Another thought I had after I raised this, is adding some sort of variable that configures the endpoints that are automatically added for the local component to use their Ingress if configured instead of the local service?

So in my case I'd enable the ingress for the compactor and the storegateway.
For my environment these are then available via a load-balancer that handles TLS termination.
Then I set the new variable from false to true (I guess it would be called something like `ingressEndpoints`?)
So if this optional var to use ingress hosts for the endpoints is set to true, instead of the `charts/thanos/templates/query/deployment.yaml`  setting `dnssrv+_grpc._tcp.*` values for `--endpoint=` parameters for the local components it instead uses the hostname configured on their ingress.
Then I add `--grpc-client-tls-secure` and any other `-grpc-client-tls-*` options I need to `query.extraArgs` to enable TLS for all endpoints.

With the above though I don't know if I hard-code the port to 443, or how to configure it.

Perhaps instead of a `ingressEndpoints` var I could have a `tlsEndpoints[]` var that if set is used for all the `--endpoint=` parameters instead of any locally templated ones?
That seems cleaner perhaps.

What do you think?

jim-barber-he:
Taking a completely different approach.

PR #73, opened 2026-04-22
feat(grpc-ingress): Create gRPC Ingress resources

jim-barber-he:

The chart has a way to expose the gRPC service via API Gateway but hasn't provided a way to do it for Ingress, so add support for that.

- fixes #

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.6.0
+version: 0.7.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.5.3](https://img.shields.io/badge/Version-0.5.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.7.0](https://img.shields.io/badge/Version-0.7.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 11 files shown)


jim-barber-he:
@hamdikh There was no way to expose the gRPC endpoints via Ingress (only API Gateway) so this PR adds that.

hamdikh (changes requested):
@jim-barber-he  thanks for the PR. A couple of points:

The chart already ships a GRPCRoute for Gateway API users, which is the recommended path going forward (ingress-nginx is [deprecated](https://kubernetes.github.io/ingress-nginx/)).
For classic Ingress users, gRPC can be exposed on the existing Ingress via a backend-protocol annotation ([nginx example](https://kubernetes.github.io/ingress-nginx/examples/grpc/)), no need for a second resource.

Could you refactor to extend the existing Ingress with an optional gRPC toggle instead? Something like:

```yaml
ingress:
  enabled: true
  grpc:
    enabled: false
    port: 10901
    annotations: {}
```
Simpler to maintain and keeps one Ingress per component.

jim-barber-he:
@hamdikh I'm using the traefik ingress controller which is not deprecated at this stage.

jim-barber-he:
> Could you refactor to extend the existing Ingress with an optional gRPC toggle instead?

Yeah I could do a grpc toggle. Is there a scenario where you'd want to run both? Perhaps not so a toggle is probably appropriate.

jim-barber-he:
@jim-barber-he Ok I have provided a `grpc` toggle on the existing ingresses instead and pushed as a separate commit.

hamdikh:
@jim-barber-he this forces to expose either grpc or http, maybe both would be better

jim-barber-he:
> this forces to expose either grpc or http, maybe both would be better

@hamdikh Ok, so I should just revert my last commit then?

hamdikh:
@jim-barber-he to rephrase what I'd like: let's keep a single Ingress template file, but with a values structure flexible enough that users can pick any ingressClass for HTTP and gRPC independently (e.g. Traefik for HTTP, Gloo for gRPC).
The template can conditionally render one or two Ingress resources depending on whether the gRPC class matches the HTTP one, but it all lives in the same file, no second template to maintain.
Values shape could be something like:

```yaml
ingress:
  enabled: true
  ingressClassName: traefik
  annotations: {}
  hosts: [...]
  tls: []
  grpc:
    enabled: false
    ingressClassName: gloo
    annotations: {}
    hosts: [...]
    tls: []
 ```
 
Could you update the PR along those lines? Keeps it simple for users while covering the split-controller case.

**Comment by jim-barber-he** [comment]
@hamdikh 

That config feels a little awkward in that if you only wanted the grpc endpoint your values might look like this for example:
```yaml
ingress:
  enabled: false
  grpc:
    enabled: true
    ingressClassName: traefik
    hosts: [...]
```

So `ingress.enabled: false` turns off http, but almost looks like it would turn off gRPC too.
The "fix" would be to use something like this:
```yaml
ingress:
  http:
    enabled: true
    ingressClassName: traefik
    annotations: {}
    hosts: [...]
    tls: []
  grpc:
    enabled: false
    ingressClassName: gloo
    annotations: {}
    hosts: [...]
    tls: []
```
But that would be a breaking change, so I'd imagine you don't want me to pursue this. Any other ideas? Or shall I just go with your proposal even if it is awkward?

jim-barber-he:
@hamdikh 

I've added a 3rd commit that is an attempt to implement the configuration as you describe.
If you're happy with it, including how `ingress.enabled: false` doesn't negate `ingress.grpc.enabled: true` then feel free to squash merge.

If not, then let me know if you want me to go back to one of the previous commits, or if I need to make more changes.

jim-barber-he:
When fixing the conflicts I decided to squash it all together and clean up the commit message to cover just the current approach.

jim-barber-he:
@hamdikh The last push was to remove the ingress-grpc.yaml files and to add code into the existing ingress.yaml files to range over the http and grpc configurations to create the ingresses as you requested. 
I also had to bump the chart version again after recent merges to the master branch.

hamdikh (approved):
LGTM 👍🏽

PR #76, opened 2026-04-23
fix(ruler): add missing grpc service

marcdavoli:

This PR adds a Kubernetes (headless) Service that exposes Thanos Ruler's GRPC endpoint on (default) port 10901. This is required for Thanos Query to be able to consume Ruler's StoreAPI.

#### Checklist

### DIFF
--- a/charts/thanos/templates/ruler/service.yaml
+++ b/charts/thanos/templates/ruler/service.yaml
@@ -23,4 +23,29 @@ spec:
     app.kubernetes.io/component: ruler
     app.kubernetes.io/instance: {{ .Release.Name }}
   type: {{ .Values.ruler.service.type }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    {{- include "thanos.annotations" (dict "Values" .Values "component" "ruler") | nindent 4 }}
+    {{- with .Values.ruler.service.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  labels:
+    {{- include "thanos.labels" . | nindent 4 }}
+    {{- with .Values.ruler.service.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  name: {{ include "thanos.compName" (list . "ruler") }}-headless
+spec:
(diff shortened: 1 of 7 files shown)


charts/thanos/templates/ruler/service.yaml
```diff
@@ -23,4 +23,29 @@ spec:
...
+    {{- include "thanos.annotations" (dict "Values" .Values "component" "ruler") | nindent 4 }}
+    {{- with .Values.ruler.service.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  labels:
+    {{- include "thanos.labels" . | nindent 4 }}
+    {{- with .Values.ruler.service.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  name: {{ include "thanos.compName" (list . "ruler") }}-headless
```
marcdavoli, line 40: Follows the same pattern as Thanos Receive: the headless service is always created alongside the regular ClusterIP service.

charts/thanos/templates/ruler/statefulset.yaml
```diff
@@ -60,6 +61,8 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.ruler.service.httpPort }}
+            - name: grpc
+              containerPort: {{ .Values.ruler.service.grpcPort }}
```
marcdavoli, line 65: This was missing.

charts/thanos/templates/ruler/statefulset.yaml
```diff
@@ -42,6 +42,7 @@ spec:
           args:
             - rule
             - --http-address=0.0.0.0:{{ .Values.ruler.service.httpPort }}
+            - --grpc-address=0.0.0.0:{{ .Values.ruler.service.grpcPort }}
```
marcdavoli, line 45: Similar pattern as http above.
The flag was already implicitly added with the default port, but now we need to explicitly include it, in case the user decides to customize the port.

hamdikh (approved):
LGTM 👍🏽

PR #83, opened 2026-04-27
fix(servicemonitor): add namespace specification

GiacomoPetacchi:
Add the namespace specification for all ServiceMonitor templates, default is current release version. It can be changed via `serviceMonitor.namespace` variable.

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.6.0
+version: 0.6.1
 appVersion: "v0.41.0"
 keywords:
   - thanos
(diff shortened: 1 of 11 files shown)


hamdikh:
Thanks for the contribution. Could you describe the specific scenario where this is needed? ServiceMonitor without an explicit namespace field defaults to `.Release.Namespace` automatically, so this PR as written is a no-op. If the goal is to support cross-namespace ServiceMonitor deployment (e.g. Prometheus Operator with a restrictive ServiceMonitorNamespaceSelector), that's a valid feature but it should be:

- Re-labeled as feat: not fix:
- Tied to a use case in the description or a linked issue
- Documented as an opt-in for the cross-namespace scenario

Otherwise this just adds a config knob without solving a real problem.

GiacomoPetacchi:
Hi @hamdikh, thanks for your suggestions. My intention was to make the Helm chart more standard and align it with the expected set of features.
I also marked this as a fix because `serviceMonitor.namespace` is already defined in `charts/thanos/values.yaml` to indicate the target creation namespace, but it is not actually implemented in the chart templates.
If that makes sense, I can reopen it in another way if you prefer.

PR #104, opened 2026-05-06
feat(ruler): Rule importer improvements

viktor-f:

This PR introduces 3 updates to the ruler:

- Adds ClusterRole and ClusterRoleBinding for ruler that allows ruler to read and watch PrometheusRules in all namespaces.
- Adds containerSecurityContext to the prometheus-rule-importer sidecar container and includes "runAsUser: 1000" in the default for this pod. This is especially important since the default alpine/kubectl container will otherwise run as root.
- Adds some minor extra output to the rule importer script. I think this makes it a little bit easier to see if things have succeeded or not.

I have not yet signed the DCO, will look into that later today if I have time. But I wanted to get the PR up in order to get feedback. I am also ok if any of you maintainers take these changes and rework them in another PR.

These three commits could be split into separate PRs as well, let me know if that is needed.

#### Checklist

### DIFF
--- a/charts/thanos/templates/ruler/statefulset.yaml
+++ b/charts/thanos/templates/ruler/statefulset.yaml
@@ -96,6 +96,7 @@ spec:
               mountPath: /generated
             - name: rule-importer-script
               mountPath: /scripts
+          {{- include "thanos.containerSC" (dict "root" . "key" "ruler") | nindent 10 }}
         {{- end }}
         {{- include "thanos.extraContainers" (dict "Values" .Values "component" "ruler" "root" .) | nindent 8 }}
       {{- include "thanos.dnsConfig" (dict "Values" .Values "component" "ruler") | nindent 6 }}
--- a/charts/thanos/values.yaml
+++ b/charts/thanos/values.yaml
@@ -2581,8 +2581,18 @@ ruler:
     fsGroupChangePolicy: OnRootMismatch
 
   # -- Container security context for Ruler. Overrides global.containerSecurityContext.
-  # @default -- {}
-  containerSecurityContext: {}
+  # Default includes "runAsUser: 1000" in order to force alpine/kubectl sidecar to not run as root
+  # @default -- global defaults plus "runAsUser: 1000"
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+      - ALL
+    readOnlyRootFilesystem: true
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
+    runAsUser: 1000
 
   # -- DNS configuration for Ruler pods. Overrides global.dnsConfig.
   # @default -- {}
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1126,7 +1126,7 @@ The table below documents all available values. Top-level keys group settings by
 | ruler.autoImportPrometheusRules.sidecar.image.pullPolicy | string | `"IfNotPresent"` | Image pull policy for the kubectl sidecar. |
 | ruler.autoImportPrometheusRules.sidecar.image.repository | string | `"alpine/kubectl"` | Repository for the kubectl sidecar that reads PrometheusRule CRDs. |
 | ruler.autoImportPrometheusRules.sidecar.image.tag | string | `"latest"` | Tag for the kubectl sidecar image. |
-| ruler.containerSecurityContext | object | {} | Container security context for Ruler. Overrides global.containerSecurityContext. |
+| ruler.containerSecurityContext | object | global defaults plus "runAsUser: 1000" | Container security context for Ruler. Overrides global.containerSecurityContext. Default includes "runAsUser: 1000" in order to force alpine/kubectl sidecar to not run as root |
 | ruler.dnsConfig | object | {} | DNS configuration for Ruler pods. Overrides global.dnsConfig. |
 | ruler.enabled | bool | `false` | Enable the Ruler StatefulSet. |
 | ruler.extraArgs | list | [] | Additional CLI arguments appended to the `thanos rule` command. |
--- a/charts/thanos/templates/ruler/configmap-auto-import-prometheus-rules.yaml
+++ b/charts/thanos/templates/ruler/configmap-auto-import-prometheus-rules.yaml
@@ -41,11 +41,15 @@ data:
         echo "Rule files changed, triggering Thanos Ruler reload"
         if curl -sf -X POST http://localhost:10902/-/reload; then
           PREV_SUM="$CURRENT_SUM"
+          echo "Reload successful"
         else
           echo "Failed to reload Thanos Ruler, will retry next cycle"
         fi
+      else
+        echo "Rules files have not changed, reload not needed"
       fi
 
+      echo "Sleeping"
       sleep 60
     done
 {{- end }}


charts/thanos/templates/ruler/statefulset.yaml
```diff
@@ -96,6 +96,7 @@ spec:
               mountPath: /generated
             - name: rule-importer-script
               mountPath: /scripts
+          {{- include "thanos.containerSC" (dict "root" . "key" "ruler") | nindent 10 }}
```
viktor-f, line 99: Ideally I would have wanted to put this as a separate key (e.g. ruler.autoImportPrometheusRules.sidecar.containerSecurityContext), but I didn't manage to get this to read a sub-key under ruler using the same helper function.
If you have suggestions on how to re-write this to allow this, then let me know. But I don't think it is too bad to have both containers use the same security context.

charts/thanos/values.yaml
```diff
@@ -2006,6 +2006,13 @@ ruler:
               annotations:
                 summary: Example alert firing
 
+  # Add a ClusterRole and ClusterRoleBinding that allows thanos ruler to read and watch PrometheusRules
+  # Also requires global.rbac.create to be true
+  # This is likely wanted if ruler.autoImportPrometheusRules is enabled
+  clusterRoleReadPrometheusRules:
```
viktor-f: Not sure this name is great, let me know if you have other suggestions.

hamdikh (commented):
Thanks for the contribution @viktor-f, this looks good to me overall. The RBAC addition and the extra output in the importer script are welcome improvements.

A couple of points though:

**On the security context**: I don't think we should set `runAsUser: 1000` (or any specific CSC) by default on behalf of users. That's a behavior change that could break existing deployments where users have their own PSP/PSA setup, custom UIDs, or volume permissions tied to a different user. I'd rather we document and encourage hardening (e.g. in the README or values comments) but leave the actual values empty by default so users opt in explicitly. The sidecar should follow the same principle.

**On the naming**: I agree `clusterRoleReadPrometheusRules` isn't ideal. Since this RBAC is only useful when `autoImportPrometheusRules` is enabled, what do you think about nesting it under that key, something like `ruler.autoImportPrometheusRules.rbac.create`? That keeps the relationship between the two options clear.

Could you please:
- Sign your commits (DCO)
- Rebase on top of master
- Address the points above
- Ping me once that's done and I'll take another pass

Thanks!

viktor-f:
> Thanks for the contribution @viktor-f, this looks good to me overall. The RBAC addition and the extra output in the importer script are welcome improvements.


Thanks for the thoughts @hamdikh, I have some more questions before doing all of the fixes. I have not done all of the things you asked of me yet, but I want your input on the security context before I finish.

> A couple of points though:
> 
> **On the security context**: I don't think we should set `runAsUser: 1000` (or any specific CSC) by default on behalf of users. That's a behavior change that could break existing deployments where users have their own PSP/PSA setup, custom UIDs, or volume permissions tied to a different user. I'd rather we document and encourage hardening (e.g. in the README or values comments) but leave the actual values empty by default so users opt in explicitly. The sidecar should follow the same principle.

I do agree with what you are saying in general. The only issue here is that without `runAsUser: 1000` the sidecar will not start due to the default security context having `runAsNonRoot: true` but the image is by default trying to run as root. So you get this error in the pod `Error: container has runAsNonRoot and image will run as root (pod: "thanos-ruler-1_thanos(98623943-90ee-4008-a99d-a78022459f1d)", container: prometheus-rule-importer)`.

So while I agree with your points in general, I think it is more important that the default values should get the pods in a running state.
It would be possible to instead have some sort of override for the sidecar so that it has `runAsNonRoot: false` and not set the user. But I would rather have the defaults be on the more secure side and set `runAsUser: 1000` (or possibly some other non-root user).
A third alternative would be to use another image that does not run as root by default (either something that is available from another source or something that is built specifically for this project). But this of course has impact on what other sources you trust or the maintainability of this project, I will not give any opinions if this is feasible and instead leave that to you as maintainer.

What do you think about this? I'm happy to further discuss these different options.

> **On the naming**: I agree `clusterRoleReadPrometheusRules` isn't ideal. Since this RBAC is only useful when `autoImportPrometheusRules` is enabled, what do you think about nesting it under that key, something like `ruler.autoImportPrometheusRules.rbac.create`? That keeps the relationship between the two options clear.

My original thought was to decouple the rbac from the sidecar. Mostly in case someone decides to have their own implementation of the sidecar and would still want the rbac added. But I realize now that it is likely a bit outside the scope of this chart. If someone wants such a custom solution, then they can also add the rbac separately. So yes, I will move the config option to `ruler.autoImportPrometheusRules.rbac.create`.

> Could you please:
> 
>     * Sign your commits (DCO)
> 
>     * Rebase on top of master
> 
>     * Address the points above
> 
>     * Ping me once that's done and I'll take another pass
> 
> 
> Thanks!

viktor-f:
I have now signed the DCO (if I understood everything correctly), rebased on main and changed the rbac config value as you suggested.

I'll wait for you to reply on my other comment before doing any other changes.

viktor-f:
@hamdikh have you had any time to look at my comments above?

viktor-f:
I noticed that the RBAC changes were already done in another PR, so I removed that commit.
@hamdikh when do you think you might have time to look at and answer the comments I made above?

hamdikh:
Thanks @viktor-f. I'm not comfortable with `ruler.containerSecurityContext` applying to both the sidecar and the main `thanos-rule` container, that's a breaking change for current users.

Could you scope this to a dedicated key like `ruler.autoImportPrometheusRules.sidecar.containerSecurityContext`? And for the `thanos.containerSC` sub-key issue you hit, please open a separate issue with the error output so it can be debugged in the open.

The script output changes are fine and could land as a small standalone PR.

Could you also open a specefic issue for this one as well with some error logs input for me to better understand what's breaking ?

viktor-f:
Thanks for your reply.

> Thanks @viktor-f. I'm not comfortable with `ruler.containerSecurityContext` applying to both the sidecar and the main `thanos-rule` container, that's a breaking change for current users.
> 
> Could you scope this to a dedicated key like `ruler.autoImportPrometheusRules.sidecar.containerSecurityContext`? And for the `thanos.containerSC` sub-key issue you hit, please open a separate issue with the error output so it can be debugged in the open.

I think it makes sense to put the security context for the sidecar into a separate key. Previously there was no support for getting security context from deeper keys in the config from what I understood, but I see now that you [reworked](https://github.com/thanos-community/helm-charts/commit/5c804e2b5b77cb88acd55a02fa80f15bce6492ff#diff-a12b3d2485d8e1ed9458f8821b969e37ec9b42cafca7bbd9a1ce34d9cd470e48R80) it to be able to get security context from two keys deep (e.g. `ruler.sidecar.securityContext`). However, the suggestion with `ruler.autoImportPrometheusRules.sidecar.containerSecurityContext` would need a new rewrite of the helper function since that is three keys deep in the config. I'll try to implement that and see if you like it, instead of writing an issue for it.
 
> The script output changes are fine and could land as a small standalone PR.

Done, https://github.com/thanos-community/helm-charts/pull/158 

> Could you also open a specefic issue for this one as well with some error logs input for me to better understand what's breaking ?


Do you here mean that you want an issue that explains the issue of running the sidecar without `runAsUser: 1000` or are you referencing something else?

PR #106, opened 2026-05-06
fix(query): Deduplicate metrics against receive_replica label

### LINKED ISSUE #105: Querier doesn't deduplicate metrics due to "receive_replica" label
## Environment

| Chart version      |  0.8.4     |
| Thanos version     |   0.41.0    |
| Helm version       |       |
| Kubernetes version |       |
| Cloud / Platform   |       |

## Values used

Default configuration

## What happened
With 2 prometheus in HA using remote write (with a unique `prometheus_replica` external label set), the metrics are still duplicated when queried.

It seems that the receiver is adding a `receive_replica` label, but the querier is not deduplicating against it

Deduplicatated metrics

## How to reproduce it

1. Send metrics via remote write from 2 prometheus in HA
2. Do a query

bigon:
The receiver is adding a receive_replica label, but the querier is not deduplicating the metrics against it.

- Fixes: #105

#### Checklist

- [X] [DCO](https://developercertificate.org) signed
- [X] Chart Version bumped (if the pr is an update to an existing chart)
- [X] Variables are documented in the README.md

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.9.0
+version: 0.9.1
 appVersion: "v0.41.0"
 keywords:
   - thanos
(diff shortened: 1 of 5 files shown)


hamdikh:
Thanks @bigon, fix looks right. Could you add a unit test in tests/query_test.yaml asserting the default args contain both --query.replica-label=prometheus_replica and --query.replica-label=receive_replica? The existing case overrides replicaLabels so it won't catch regressions on the default.

bigon:
@hamdikh I've added a test and rebased

hamdikh (approved):
LGTM 👍🏽

PR #122, opened 2026-05-09
feat(receive): add split mode (router + ingester)

### LINKED ISSUE #54: feature: support splitting Routing Receivers from Ingesting Receivers
## Problem

This is in reference to this proposal: https://thanos.io/tip/proposals-accepted/202012-receive-split.md.

The chart supports creating Receivers using the `receiver.*` values, but the chart does not support creating separate "Routing" Receivers (AKA distributor mode) & "Ingesting" Receivers. 

Other Thanos helm charts already support this split, see:
- Separate `ingestor` and `router` keys in https://github.com/stevehipwell/helm-charts/blob/main/charts/thanos/values.yaml#L576-L742
- Separate `receive` and `receiveDistributor` keys in https://github.com/bitnami/charts/blob/main/bitnami/thanos/values.yaml#L4675

## Proposed solution

Implement the ability to create single-purpose Receivers, possibly using a `mode` field in the `receiver` block, or using separate blocks altogether, like
- `receiver`: Does both Routing and Ingesting
- `receiverIngest`: Does only Ingesting
- `receiverRouting`: Does only Routing

## Alternatives considered

Patching the helm chart for this purpose quickly becomes too messy to maintain.

## Additional context

Supporting this feature is necessary for organizations that require high availability and scalability of the Receive write path.

hamdikh:
#### What this PR does / why we need it

Adds a new `receive.mode` value to support deploying Thanos Receive as separate Router and Ingester workloads, following the [receive split proposal](https://thanos.io/tip/proposals-accepted/202012-receive-split.md). This is necessary for organizations that need high availability and horizontal scalability of the remote-write path.

- `receive.mode: standalone` (default) — single combined RouterIngestor StatefulSet. **Fully backwards-compatible: rendering byte-identical to the previous chart.**
- `receive.mode: split` — stateless `receive-router` Deployment in front of a stateful `receive-ingester` StatefulSet. The router terminates client `remote_write`, the ingester writes to the local TSDB WAL and ships blocks to the object store.

In split mode the existing `receive.*` block configures the Ingester StatefulSet (renamed `<release>-thanos-receive-ingester`) and a new `receive.router.*` block configures the Router (`<release>-thanos-receive-router`). Router-only flags (`--receive.hashrings-file`, `--receive.hashrings-algorithm`, `--receive.local-endpoint`) and the hashrings configmap mount are stripped from the ingester. The hashrings configmap content auto-generates against ingester pod DNS in split mode.

To minimise duplication, the existing template helpers (`thanos.affinity`, `thanos.nodeSelector`, `thanos.tolerations`, `thanos.priorityClassName`, `thanos.topologySpreadConstraints`, `thanos.dnsConfig`, `thanos.podSC`, `thanos.containerSC`, `thanos.httpProbes`, `thanos.extraEnvItems` / Block, `thanos.extraEnvFromItems` / Block, `thanos.initContainers`, `thanos.extraContainers`, `thanos.extraVolumeItems` / Block, `thanos.extraMountItems` / Block) gained an optional `parent` argument. When set, the helper resolves with `component → parent → global` precedence — letting `receive.router.*` cleanly inherit anything not explicitly overridden from `receive.*`. The change is fully backwards-compatible: helpers called without `parent` behave exactly as before.

- fixes #54

#### Special notes for your reviewer

- Standalone-mode rendering is unchanged. A diff of \`helm template . --set global.objstore.config=dummy\` against \`master\` is empty for all standalone resources.
- The schema rejects invalid mode values: \`helm template . --set receive.mode=bogus\` fails with \`receive.mode must be one of "standalone", "split"\`.
- 22 new unit tests cover both modes and the new helpers (89 total receive tests, 555 total chart tests, all green).
- Chart version bumped \`0.10.1\` → \`0.11.0\`.
- README regenerated and a Split mode usage section added under \`### Receive\`.
- Migration: switching an existing release from \`standalone\` → \`split\` renames the StatefulSet (\`-receive\` → \`-receive-ingester\`), so PVCs need to be migrated manually if the user wants to retain the local TSDB WAL data.

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.10.1
+version: 0.11.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.10.1](https://img.shields.io/badge/Version-0.10.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.11.0](https://img.shields.io/badge/Version-0.11.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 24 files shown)


hamdikh:
@marcdavoli this implements the split-mode you proposed in #54. Would appreciate your review when you have a moment, especially on the API shape (`receive.mode` + nested `receive.router.*` block) and whether the parent-fallback pattern feels right for your use case.

marcdavoli:
This would work, but I have 2 questions/concerns:
1. The semantics of the `receive.router` field are a little unintuitive to me. For example, if the user doesn't set `receive.mode: split` and configures `receive.router` they won't see what they expect. There is a comment that explain this, but I'd be worried that users may fall into that trap. Maybe that's just me being overcautious though.
2. I assume the parent-fallback pattern was a necessity, because of the new `receive.router` child field? I find that too heavy of a complexity cost to pay, just for the convenience of having all the receive (split/standalone) configs being configurable in a single block. IMO, this further makes me lean towards thinking we should have a separate entire block. Maybe having 3 blocks, like: 

- receiver: Does both Routing and Ingesting
- receiverIngest: Does only Ingesting
- receiverRouting: Does only Routing

with validation rules that enforce that you can only use the first one or the last two, but never at the same time?

hamdikh:
Hey @marcdavoli ,
Fair concerns, but I'd lean toward keeping things consolidated. I can restructure the values like this:
```yaml
receive:
  enabled: true
  mode: split  # or "standalone"
  router:
    # router-specific config
  ingester:
    # ingester-specific config
```

Each part gets its own sub-object, so the config is explicit and scoped, but we keep a single top-level receive block to avoid duplicating shared knobs across separate blocks.
I'd also lean toward a single config for the ingester since it'll be enabled by default in both modes, so no need to gate it behind extra structure.
I'd prefer this over the 3-block split to keep the base code minimal and the config DRY. 
WDYT?

marcdavoli:
I think that could be a good middle ground 👍 

However, about this
> I'd also lean toward a single config for the ingester since it'll be enabled by default in both modes, so no need to gate it behind extra structure.

I think we should flesh out the details of which fields are populated in different modes.

Here is my understanding:
- In Split mode:
  - Both children fields `receive.router` and `receive.ingester` must be populated
  - The parent `receive` field can only declare certain fields (e.g.: `enabled`)?
- In Standalone mode:
  - Only `receive.ingester` must be populated
  - The parent `receive` field can only declare certain fields (e.g.: `enabled`)?
  - The child `receive.router` must be empty?

hamdikh:
Not quite @marcdavoli  let me clarify the model a bit.

In Split mode, only the child fields (receive.router and receive.ingester) should be populated. The top-level receive fields are effectively ignored or unused in that mode.

In Standalone mode, it's the opposite: all configuration lives directly under receive, and the child fields (receive.router, receive.ingester) should be left empty no need to touch them.

So the child fields are optional overall, but become required when mode is set to split. That constraint would be enforced via values schema validation.

Does that match your mental model? 

PS: This also has the benefit of preserving backward compatibility, minimizing the impact on users who stick with the default standalone mode.

marcdavoli:
That sounds good to me!

hamdikh:
@marcdavoli I'll let you take a final look then before merging on my side :)

Alphadelta14:
in practice, it's pretty challenging to adopt because it requires fully setting fields like pdb and vpa (which may not even be enabled) and probes which should have reasonable defaults.

PR #138, opened 2026-05-17
fix(ruler): Add objstore config and auto-generated Query endpoint for Ruler

### LINKED ISSUE #136: Ruler StatefulSet does not mount or pass object storage configuration
## Environment

| Chart version      | 0.11.0      |
| Thanos version     | v0.41.0      |

## What happened

The `ruler` StatefulSet rendered by this chart does not configure object storage for `thanos rule`.

The chart has a global object storage configuration under `global.objstore`, and other stateful components such as Store Gateway, Compactor, Receive, and Bucket pass it to Thanos via:

```yaml
--objstore.config-file=/etc/thanos/objstore.yml
```

and mount the object store Secret into `/etc/thanos`.

However, the Ruler StatefulSet only includes:

```yaml
- rule
- --http-address=...
- --grpc-address=...
- --data-dir=/var/thanos/ruler
- --rule-file=/etc/thanos/rules/*.yaml
- --alertmanagers.config-file=/etc/thanos/alertmanagers.yaml
- --query.config-file=/etc/thanos/query-urls.yaml
```

It also adds `checksum/objstore-configuration`, but the object store Secret is not mounted and `--objstore.config-file` is not passed to the container.

## Why this matters

Thanos Ruler needs object storage when recording rules are used and the generated time series should be persisted as TSDB blocks and later queried through Thanos Query/StoreAPI.

The upstream Thanos Rule documentation states that, by default, rule evaluation results are written to local TSDB, Ruler exposes StoreAPI, and generated TSDB blocks are uploaded to object storage. The documented `thanos rule` example includes:

```bash
--objstore.config-file "bucket.yml"
```

Without this configuration, recording rule results cannot be uploaded to object storage by the Ruler.

When `ruler.enabled=true`, the chart should support object storage configuration for the Ruler, likely by doing the same as other object-store-backed components:

- add `--objstore.config-file=/etc/thanos/objstore.yml` to the Ruler args
- mount the `global.objstore.secretName` Secret into the Ruler pod
- read `global.objstore.secretKey` as `objstore.yml`

Optionally, this could be controlled by a Ruler-specific value if maintainers want to support alerting-only Ruler deployments without requiring object storage.

## Actual behavior

The Ruler StatefulSet does not mount the object storage Secret and does not pass `--objstore.config-file` to `thanos rule`.

r0bj:

This updates the Ruler chart templates so `thanos rule` is configured consistently with the other object-store-backed components.

## Changes

- Pass `--objstore.config-file=/etc/thanos/objstore.yml` to the Ruler container.
- Mount the global objstore Secret into the Ruler pod using `global.objstore.secretName` and `global.objstore.secretKey`.
- Add Ruler as an auto-generated Query StoreAPI endpoint when `query.endpoints.autogen.enabled=true`.
- Update unit tests and values documentation for the new behavior.

Fixes #136.

@hamdikh please review.

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.11.0
+version: 0.11.1
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -751,7 +751,7 @@ The table below documents all available values. Top-level keys group settings by
 | query.containerSecurityContext | object | {} | Container security context for Query. Overrides global.containerSecurityContext. |
 | query.dnsConfig | object | {} | DNS configuration for Query pods. Overrides global.dnsConfig. |
 | query.enabled | bool | `true` | Enable the Query Deployment. |
-| query.endpoints.autogen.enabled | bool | `true` | Auto-generate endpoint arguments for the in-chart components (Receive, Store Gateway). These use the dnssrv+_grpc._tcp.<svc>.<namespace>.svc.<cluster-domain> format. |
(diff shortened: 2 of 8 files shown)


hamdikh:
Superseded by #155, which carries these Ruler objstore changes forward (fixes #136) and additionally fixes the Store Gateway caching-bucket mount conflict (#152). Thanks for the original work here @r0bj 🙏

PR #139, opened 2026-05-18
feat(alerting): add disabledAlerts & alertOverrides spec

### LINKED ISSUE #129: feat: allow disabling individual built-in PrometheusRule alerts
## Problem

The Thanos Helm chart currently allows enabling or disabling the built-in `PrometheusRule` resource as a whole via `global.thanosRules.enabled`, and it also supports enabling or disabling complete rule groups via `global.thanosRules.groups.<group>.enabled`.

This is useful in many cases, but it is too coarse-grained when only a single built-in alert needs to be disabled.

In real deployments, some alerts may be too noisy, too sensitive for a specific environment, or require a small local adjustment to the expression, labels, annotations, or `for` duration. Today, the only practical options are to keep the noisy alert enabled, disable the entire group, or stop using the built-in rules and maintain a custom rule set externally.

## Proposed solution

Add support for disabling individual alerts by alert name while keeping the rest of the built-in rule group enabled.

For example:

```yaml
global:
  thanosRules:
    groups:
      thanosQuery:
        enabled: true
        disabledAlerts:
          - ThanosQueryOverload
          - ThanosQueryHttpRequestQueryErrorRateHigh
```

or alternatively as a global map:

```yaml
global:
  thanosRules:
    disabledAlerts:
      thanos-query:
        - ThanosQueryOverload
        - ThanosQueryHttpRequestQueryErrorRateHigh
```

This would allow users to disable a specific built-in alert and provide their own replacement through an additional `PrometheusRule`, without losing the rest of the built-in alert coverage for that component.

### Use case

For example, an operator may want to keep all built-in `thanos-query` alerts, but replace only `ThanosQueryOverload` with a slightly different threshold or `for` duration that better matches their production traffic pattern.

With per-alert disabling, they could disable the built-in alert and deploy a custom replacement with the same alert name or a locally adjusted name, while still receiving future updates for the rest of the chart-managed rules.

### Benefits

- Avoids disabling an entire rule group because of one noisy alert.
- Makes the built-in rules easier to adopt in production environments.
- Allows local customization without fully forking or replacing the chart-managed rules.
- Keeps the existing group-level enable/disable behavior intact.

### Backward compatibility

This can be added as an opt-in feature with an empty default list/map, so existing installations would continue rendering the same rules by default.

trouaux:
adding disabledAlerts and alertOverrides feature

- fixes #129 

#### Checklist

### DIFF
--- a/charts/thanos/templates/prometheusrule/_helpers.tpl
+++ b/charts/thanos/templates/prometheusrule/_helpers.tpl
@@ -145,8 +145,7 @@ Usage:
 {{- define "thanos.rules.processGroups" -}}
 {{- $ctx := index . 0 -}}
 {{- $tr := index . 1 -}}
-{{- $disabledAlerts := $tr.disabledAlerts | default list }}
-{{- $rulesYaml := printf "groups:\n%s%s%s%s%s%s%s%s"
+{{- $allRules := list
   (include "thanos.rules.compact"         $ctx)
   (include "thanos.rules.query"           $ctx)
   (include "thanos.rules.receive"         $ctx)
@@ -156,23 +155,25 @@ Usage:
   (include "thanos.rules.bucketReplicate" $ctx)
   (include "thanos.rules.componentAbsent" $ctx)
 -}}
+{{- $rulesYaml := printf "groups:\n%s" (join "\n" $allRules) -}}
 {{- $parsedRules := $rulesYaml | fromYaml -}}
-{{- range $parsedRules.groups }}
-{{- $group := . }}
-{{- $filteredRules := list }}
-{{- range $group.rules }}
-{{- if not (has .alert $disabledAlerts) }}
-{{- $rule := . }}
-{{- if and $tr.alertOverrides (hasKey $tr.alertOverrides .alert) }}
-{{- $overrides := index $tr.alertOverrides .alert }}
-{{- $mergedLabels := merge (deepCopy $overrides) $rule.labels }}
-{{- $rule = merge (dict "labels" $mergedLabels) $rule }}
-{{- end }}
-{{- $filteredRules = append $filteredRules $rule }}
-{{- end }}
-{{- end }}
+{{- $disabledAlerts := $tr.disabledAlerts | default list -}}
+{{- range $parsedRules.groups -}}
+{{- $group := . -}}
+{{- $filteredRules := list -}}
+{{- range $group.rules -}}
+{{- if not (has .alert $disabledAlerts) -}}
+{{- $rule := . -}}
+{{- if and $tr.alertOverrides (hasKey $tr.alertOverrides .alert) -}}
+{{- $overrides := index $tr.alertOverrides .alert -}}
+{{- $mergedLabels := merge (deepCopy $overrides) $rule.labels -}}
+{{- $rule = merge (dict "labels" $mergedLabels) $rule -}}
+{{- end -}}
+{{- $filteredRules = append $filteredRules $rule -}}
+{{- end -}}
+{{- end -}}
 {{- if $filteredRules }}
 - {{ merge (dict "rules" $filteredRules) (omit $group "rules") | toYaml | nindent 2 | trimSuffix "\n" }}
-{{- end }}
-{{- end }}
+{{- end -}}
+{{- end -}}
 {{- end -}}
--- a/charts/thanos/templates/prometheusrule/_helpers.tpl
+++ b/charts/thanos/templates/prometheusrule/_helpers.tpl
@@ -145,11 +145,11 @@ Usage:
 {{- define "thanos.rules.filterGroups" -}}
 {{- $parsedRules := index . 0 -}}
 {{- $tr := index . 1 -}}
+{{- $disabledAlerts := $tr.disabledAlerts | default list }}
 {{- range $parsedRules.groups }}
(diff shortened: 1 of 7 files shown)


hamdikh:
@trouaux Thank you for your contribution  !
Could you please reword your commit and follow conv commit ? 
Also can you add the tests to make sure that the feature is gated :)

hamdikh (commented):
let's simplify the rule.yaml and pre calculate the parsed rules in a helper :)

otherwise this looks good thanks for the contib :D

charts/thanos/templates/prometheusrule/_helpers.tpl
```diff
@@ -130,3 +130,38 @@ Usage:
...
+  {{- include "thanos.rules.filterGroups" (list $parsedRules $tr) | nindent 2 }}
+*/}}
+{{- define "thanos.rules.filterGroups" -}}
+{{- $parsedRules := index . 0 -}}
+{{- $tr := index . 1 -}}
+{{- range $parsedRules.groups }}
+{{- $group := . }}
+{{- $filteredRules := list }}
+{{- range $group.rules }}
+{{- $disabledAlerts := $tr.disabledAlerts | default list }}
```
hamdikh: {{- $disabledAlerts := $tr.disabledAlerts | default list }}  => {{/* recomputed every rule */}} can be a bit optimized by moving the before the loop

charts/thanos/templates/prometheusrule/rule.yaml
```diff
@@ -27,13 +27,17 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- $rulesYaml := printf "groups:\n%s%s%s%s%s%s%s%s"
```
hamdikh: The string assembly (printf + fromYaml) in rule.yaml should move entirely into thanos.rules.filterGroups. 
Right now rule.yaml owns part of the pipeline (building and parsing the YAML string) while the helper owns the other part (filtering).
This split is inconsistent the helper should own the full pipeline from raw includes to filtered output.

rule.yaml should only call:
```yaml
groups:
{{- include "thanos.rules.filterGroups" (list $ctx $tr) | nindent 2 }}
```
And filterGroups receives $ctx and $tr directly, handling the includes, string assembly, parsing, error guard, and filtering internally. 
That way rule.yaml stays purely declarative and the helper is self-contained.

charts/thanos/templates/prometheusrule/_helpers.tpl
```diff
@@ -132,20 +132,31 @@ Usage:
...
-  - parsedRules: the parsed YAML structure containing groups
+  - ctx: the context dict from thanos.rules.context
   - tr: the .Values.global.thanosRules configuration
 
 Returns the filtered groups as YAML.
 
 Usage:
-  {{- include "thanos.rules.filterGroups" (list $parsedRules $tr) | nindent 2 }}
+  {{- include "thanos.rules.processGroups" (list $ctx $tr) | nindent 2 }}
 */}}
```
hamdikh, line 144: I propose something like this instead 

```
{{- define "thanos.rules.filterGroups" -}}
{{- $ctx := index . 0 -}}
{{- $tr := index . 1 -}}
{{- $allRules := list
  (include "thanos.rules.compact"         $ctx)
  (include "thanos.rules.query"           $ctx)
  (include "thanos.rules.receive"         $ctx)
  (include "thanos.rules.sidecar"         $ctx)
  (include "thanos.rules.store"           $ctx)
  (include "thanos.rules.rule"            $ctx)
  (include "thanos.rules.bucketReplicate" $ctx)
  (include "thanos.rules.componentAbsent" $ctx)
-}}
{{- $rulesYaml := printf "groups:\n%s" (join "\n" $allRules) -}}
{{- $parsedRules := $rulesYaml | fromYaml -}}
{{- $disabledAlerts := $tr.disabledAlerts | default list -}}
{{- range $parsedRules.groups -}}
{{- $group := . -}}
{{- $filteredRules := list -}}
{{- range $group.rules -}}
{{- if not (has .alert $disabledAlerts) -}}
{{- $rule := . -}}
{{- if and $tr.alertOverrides (hasKey $tr.alertOverrides .alert) -}}
{{- $overrides := index $tr.alertOverrides .alert -}}
{{- $mergedLabels := merge (deepCopy $overrides) $rule.labels -}}
{{- $rule = merge (dict "labels" $mergedLabels) $rule -}}
{{- end -}}
{{- $filteredRules = append $filteredRules $rule -}}
{{- end -}}
{{- end -}}
{{- if $filteredRules }}
- {{ merge (dict "rules" $filteredRules) (omit $group "rules") | toYaml | nindent 2 | trimSuffix "\n" }}
{{- end -}}
{{- end -}}
{{- end -}}
```

hamdikh:
@trouaux Almost there! Two small things before we merge:

- README.md is not in the diff but the checklist has it ticked, could you add the `disabledAlerts` and `alertOverrides` params to the parameters table? or generate them using helm-docs :)
- A couple of extra test cases would be nice to have: one for a non-existent alert name in `disabledAlerts` (should be a silent no-op) and one for the default empty values (`disabledAlerts: []`, no `alertOverrides`) to make sure backward compat is explicit in the test suite.

Otherwise this looks good to go, great contribution!

PR #146, opened 2026-05-20
feat(storegateway): add sharded topology with hash and time partitioning

### LINKED ISSUE #128: feat: support sharded Store Gateway deployments
## Problem

It would be useful to add native Store Gateway sharding support to this chart, similar to the approach used by the Bitnami Thanos chart:
- https://github.com/bitnami/charts/blob/main/bitnami/thanos/values.yaml
- https://github.com/bitnami/charts/blob/main/bitnami/thanos/templates/storegateway/statefulset-sharded.yaml

The current chart deploys Store Gateway as a single StatefulSet scaled with `storegateway.replicaCount`. This provides replica-level availability, but every Store Gateway replica is still responsible for the same full object storage block set.

For larger Thanos installations, this can become a bottleneck. Store Gateway queries that touch many historical blocks can be expensive in terms of CPU, memory, disk cache, and object storage access. As the number of blocks grows, simply increasing replicas does not reduce the per-replica block responsibility.

## Proposed solution

Add optional Store Gateway sharding support under a value such as:

```yaml
storegateway:
  sharded:
    enabled: false
    hashPartitioning:
      shards: 3
      extraRelabelingConfigs: []
    timePartitioning: []
```

When enabled, the chart could render multiple Store Gateway StatefulSets and Services, where each shard is responsible for only part of the object storage block set.

Possible supported strategies:

- hash-based partitioning using `--selector.relabel-config` with `hashmod` on `__block_id`
- time-based partitioning using `--min-time` and `--max-time`
- optionally combining both approaches

This matches Thanos’ documented sharding model for object-store readers and Store Gateway.

### Expected behavior

Users should be able to deploy multiple Store Gateway shards from Helm values, without manually duplicating chart releases or maintaining custom templates.

The Query component should be able to discover all Store Gateway shards automatically, for example through generated endpoints or DNS discovery, so that queries still cover the full historical data set.

### Benefits

- Allows Store Gateway to scale horizontally for larger object storage workloads
- Reduces the amount of block metadata/cache each Store Gateway instance needs to handle
- Improves operational ergonomics for large Thanos deployments
- Aligns this chart with existing Thanos sharding capabilities
- Avoids users having to maintain custom chart overlays for sharded Store Gateway setups

### Alternatives considered

The current workaround is to deploy multiple Store Gateway instances manually with custom `extraArgs`, separate Services, and manual Query endpoint configuration. This works, but it is error-prone and hard to maintain as the number of shards grows.

### Additional context

Thanos documents Store Gateway sharding through relabeling and time partitioning:

https://thanos.io/v0.40/thanos/sharding.md/

Bitnami’s Thanos chart already exposes a similar concept via `storegateway.sharded`, which could be used as implementation inspiration, though the implementation should be adapted to this chart’s existing naming, Service discovery, tests, and values structure.

hamdikh:

Adds an opt-in **sharded Store Gateway topology** to the chart. When
`storegateway.sharded.enabled=true`, the chart renders one StatefulSet per
shard (`<fullname>-storegateway-<index>`) so each shard serves a subset of the
object-storage blocks instead of every replica loading the full block set.

Shards are the cross-product of:

- `storegateway.sharded.timePartitioning` — a list of `{ min, max }` ranges,
  each rendered as `--min-time` / `--max-time`.
- `storegateway.sharded.hashPartitioning.shards` — a `hashmod` on `__block_id`
  rendered into `--selector.relabel-config` (optionally prefixed with
  `hashPartitioning.extraRelabelingConfigs`).

Either dimension can be used on its own. `replicaCount` becomes the replica
count *per shard*.

Discovery and supporting resources:

- The aggregate Service becomes **headless** in sharded mode so Query's
  `dnssrv+_grpc._tcp` discovery resolves every shard pod and fans out across
  all shards — no Query config change required.
- Per-shard HTTP Services are added for addressing individual shards (UI,
  `/metrics`, debug). They omit the `component` label so the shared
  ServiceMonitor scrapes only the aggregate Service (no double-scrape).
- HPA, PDB and the shard selector labels are shard-aware; the caching ConfigMap
  and NetworkPolicy remain shared across shards.

**Non-sharded output is unchanged** (single unsuffixed StatefulSet, no shard
labels), so this is fully backward compatible.

- fixes #128

- New helper `thanos.storegateway.shards` in `_helpers.tpl` computes the shard
  list and is consumed by the statefulset/service/hpa/pdb templates.
- `values.schema.json` was updated by hand (the locally available `helm schema`
  plugin version no longer reproduces the committed schema), matching the
  existing style.
- Tested with `helm lint --strict` and `helm unittest .` (612 tests pass,
  including new sharding cases). Rendered sharded/non-sharded output verified.

#### Checklist

- [x] Chart Version bumped (0.13.0 → 0.14.0)


hamdikh:
@r0bj this implements the sharded Store Gateway you requested in #128 would you mind taking a look ?

Summary of the approach:

- Opt-in via `storegateway.sharded.enabled`. One StatefulSet per shard, named `<fullname>-storegateway-<index>`.
- Two partitioning dimensions, usable together (cross-product) or independently:
  - **Hash**: `sharded.hashPartitioning.shards` → `hashmod` on `__block_id` via `--selector.relabel-config` (with optional `extraRelabelingConfigs` prefix).
  - **Time**: `sharded.timePartitioning` list of `{ min, max }` → `--min-time` / `--max-time`.
- `replicaCount` is now per-shard.
- The aggregate Service is made **headless** in sharded mode so Query's existing `dnssrv+_grpc._tcp` discovery fans out across all shard pods with no extra config. Per-shard HTTP Services are added for addressing individual shards; the ServiceMonitor still targets only the aggregate Service (no double-scrape).
- Non-sharded rendering is unchanged (fully backward compatible).

A couple of things I'd especially value your eyes on:
1. The shard naming scheme / cross-product index ordering note that changing shard count or time partitions renumbers shards and triggers StatefulSet recreation.
2. Whether the default `hashPartitioning.shards: 0` (meaning "hash partitioning off") is intuitive vs. defaulting to `1`.

r0bj:
Thanks for the PR, the approach looks reasonable to me.

I compared it with the Bitnami Thanos chart, which I used before for Store Gateway sharding, and the overall model is similar: one StatefulSet per shard with hash/time partitioning support.

Regarding `hashPartitioning.shards`, I think defaulting to `1` would be more intuitive than `0`. Conceptually, disabled hash partitioning is equivalent to having a single hash shard, so `1` feels like the more natural default.

PR #164, opened 2026-05-26
feat(thanos): support existingClaim for all statefulset component persistence

Alja9:

Adds `persistence.existingClaim` to every StatefulSet component: `compactor`, `storegateway`, `ruler`, `receive` (standalone) and `receive.ingester` (split mode).

When set, the named PVC is mounted directly as the `data` volume and no `volumeClaimTemplate` is generated, which is what you want when the volume is managed outside the chart — restored from a snapshot, provisioned by Terraform, or shared with another tool. Defaults to `""`, so existing installs render exactly as before.

**Store Gateway sharding.** The `volumeClaimTemplates` guard is `(or $shard.isSharded (not ...existingClaim))`, so a sharded Store Gateway falls back to generated claims and ignores `existingClaim`; a single pre-provisioned PVC cannot be split across shards. Covered by a sharded test that asserts every shard still gets a `volumeClaimTemplate` and that no shard mounts the claim.

**Split mode.** In split mode persistence resolves from `receive.ingester.persistence`, not `receive.persistence`, so `receive.ingester.persistence` now carries real chart defaults (`enabled`, `existingClaim`, `size`, `storageClass`, `volumeAttributesClassName`, `accessModes`) instead of `ingester: {}`. This also satisfies the `persistence` entry that the schema's `mode: split` `allOf` already required, so a split-mode install no longer has to spell the block out. `_split_defaults.yaml` and the values-table docs are updated to match.

**Receive accessModes.** `receive/statefulset.yaml` hardcoded `accessModes: ["ReadWriteOnce"]`, silently ignoring `receive.persistence.accessModes`, which every other component honours. Fixed here because `existingClaim` is documented as taking precedence over `accessModes` and that has to mean something. Covered by its own test.

**values.schema.json.** `existingClaim` (`string`, default `""`) added under `compactor.persistence`, `receive.persistence`, `receive.ingester.persistence`, `ruler.persistence` and `storegateway.persistence`.

**Scope.** Query, Query Frontend and BucketWeb are stateless Deployments with no persistence block and no PVC — BucketWeb reads blocks straight from the object store on each request — so there is nothing for `existingClaim` to attach to. The five PVC-producing StatefulSet components are the full set.

**Docs.** A `Persistence` section in `README.md.gotmpl` covers generated claims, `existingClaim`, and the three caveats: the claim is mounted by *every* replica (unlike a `volumeClaimTemplate`), `persistence.enabled: false` wins, and sharded Store Gateway falls back to templates.

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.23.1
+version: 0.24.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.36.0
+version: 0.37.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.23.0](https://img.shields.io/badge/Version-0.23.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.24.0](https://img.shields.io/badge/Version-0.24.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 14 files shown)


Alja9:
Hi @hamdikh ,

Could you take a look at this when you have a moment? thank you 🙏

hamdikh:
Could you rebase @Alja9  

I'll take a look at this asap

Alja9:
Done rebase

hamdikh:
@Alja9 Thanks for this, the test coverage on the non-sharded paths is solid. A few things on split mode, the values schema, and scope before it goes in.

**Split mode (receive ingester)**

In split mode the receive template resolves persistence from `receive.ingester.persistence`, not `receive.persistence`. Right now `_split_defaults.yaml` only adds `existingClaim: ""` under `receive.persistence`, so the ingester path has no explicit default. Could you:

- add `existingClaim: ""` under `receive.ingester.persistence` in `values.yaml`, and
- confirm the standalone `receive.persistence.existingClaim` default is set in `values.yaml` too (not only in the test fixture).

The split-mode tests pass because `set:` injects the value, but a plain `helm template` in split mode would fall back to nil without a real default, which we want to avoid.

**values schema**

The new keys are not in `values.schema.json` yet. Please add `existingClaim` (type `string`, default `""`) under each of:

- `compactor.persistence`
- `receive.persistence`
- `receive.ingester.persistence`
- `ruler.persistence`
- `storegateway.persistence`

Without this, schema validation will either reject the new keys or skip validating them depending on `additionalProperties`. The README rows look good, just add the `receive.ingester.persistence.existingClaim` row as well (via helm-docs) so the ingester path is documented.

**Scope**

The title scopes this to statefulset components, which makes sense. Just to make the boundary explicit for future readers: can you confirm the stateless components (Query, Query Frontend, BucketWeb) were intentionally left out ?

Alja9:
Hi @hamdikh 

Sorry for the late reply, and thanks for checking this PR.
I'll revise for the **Split mode (receive ingester)** and **values schema**.

For your question about **scope**, yes that exclusion was intentional.
Query, query frontend, and bucketweb are all stateless, those not running in statefulset (those components in the templates using `deployment.yaml`) and don't mount PVCs, no persistence block in `values.yaml` at all, example :
https://github.com/thanos-community/helm-charts/blob/c5f5072690773301b7cf3cc8d4df8572011c06d7/charts/thanos/values.yaml#L605-L607

AFAIK, those are not necessary to create PVCs, because query and query frontend hold on local data (they fan out to store API endpoints at query time), and for bucketweb reads directly blocks from object storage on each request rather than caching disk. So, there is no `existingClaim` path to add for them.
So, this PR scope is correctly limited to the five components that actually provision PVCs and running in statefulset: compactor, receive, ingester, ruler, and storegateway.

Alja9:
Done, fixed.
I've addressed all and ready for review again.

hamdikh:
Thanks for this @Alja9 , the feature is useful and the per component tests are nice to see. One blocking issue and a couple of asks before merge, all scoped to `existingClaim`.

## Blocker: storegateway breaks with sharding + existingClaim

In `templates/storegateway/statefulset.yaml` the two guards are not complementary:

```yaml
# PVC volume (only when NOT sharded)
{{- if and $.Values.storegateway.persistence.enabled $.Values.storegateway.persistence.existingClaim (not $shard.isSharded) }}
# volumeClaimTemplates (only when existingClaim is empty)
{{- if and $.Values.storegateway.persistence.enabled (not $.Values.storegateway.persistence.existingClaim) }}
```

With `persistence.enabled: true`, `existingClaim` set, and sharding on, neither branch renders: the PVC volume is skipped because the shard is sharded, and `volumeClaimTemplates` is skipped because `existingClaim` is non empty. The container still mounts a volume named `data`, so the StatefulSet references a `data` volume that does not exist and the API server rejects it. This is exactly the sharded case the README says should fall back to templates.

Fix by making the templates guard fall back in sharded mode:

```yaml
{{- if and $.Values.storegateway.persistence.enabled (or $shard.isSharded (not $.Values.storegateway.persistence.existingClaim)) }}
volumeClaimTemplates:
```

The compactor, ruler, and receive paths are correct, this only affects storegateway.

## Please add a sharded test

The storegateway tests only cover the non sharded default, so the case above is uncovered. Could you add one that enables sharding (same way the existing sharded fixtures do) with `existingClaim` set, and asserts `spec.volumeClaimTemplates` is not empty and that no `data` PVC volume referencing the claim is present. That locks in the fallback.

Happy to re review as soon as the storegateway guard and test are in. Thanks again.

Alja9:
Done, changed and added

PR #169, opened 2026-05-29
feat(thanos): add volume attributes class name for PVCs

loganmzz:

At least on EKS (AWS), [Volume Attributes classes](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/) might be used instead of Storage Class to attach metadata (e.g. EBS tags) to the PVC.

N/A

N/A

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.22.0
+version: 0.23.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.22.0](https://img.shields.io/badge/Version-0.22.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.23.0](https://img.shields.io/badge/Version-0.23.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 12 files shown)


hamdikh (changes requested):
@loganmzz thanks for the PR. Before we can merge, a few things need to be addressed:
1. Verification

Can you confirm you tested this on a cluster where VolumeAttributesClass is actually available? Note that this is GA only in K8s 1.34, and beta (feature gate required) in 1.31-1.33.
2. Add capability gating

Since not all clusters will have this API available, please gate the resources using capabilities:

```yaml
{{- if .Capabilities.APIVersions.Has "storage.k8s.io/v1/VolumeAttributesClass" }}
  # VAC resources here
{{- end }}
```

This keeps the chart compatible with older Kubernetes versions instead of failing at deploy time.

3. Documentation

Please add documentation for this feature in the README/values.yaml comments, including:

What the feature does
Minimum Kubernetes version requirements (1.34 GA, 1.31+ with feature gate)
How to enable/configure it

4. Fix the chart version

You bumped the version as a fix, but this is a new feature, so it should be a minor version bump per semver.

Once these are addressed, I'll do another review pass.

loganmzz:
@hamdikh 

> 1. Verification
> 
> Can you confirm you tested this on a cluster where VolumeAttributesClass is actually available? Note that this is GA only in K8s 1.34, and beta (feature gate required) in 1.31-1.33.

It has been tested against AWS EKS 1.35. But I can also try using a cluster provided by Docker For Windows. 

>2. Add capability gating
> 
> Since not all clusters will have this API available, please gate the resources using capabilities:
> 
> ```yaml
> {{- if .Capabilities.APIVersions.Has "storage.k8s.io/v1/VolumeAttributesClass" }}
>   # VAC resources here
> {{- end }}
> ```
> 
> This keeps the chart compatible with older Kubernetes versions instead of failing at deploy time.

As long as option isn't used/provided (and it shouldn't if not compatible) it stays compatible. To keep capabilities, I can might it failed instead if provided but not compatible (but still a mess to use with `helm template` :( )

> 
> 3. Documentation
> 
> Please add documentation for this feature in the README/values.yaml comments, including:
> 
> What the feature does Minimum Kubernetes version requirements (1.34 GA, 1.31+ with feature gate) How to enable/configure it

👍 

> 4. Fix the chart version
> 
> You bumped the version as a fix, but this is a new feature, so it should be a minor version bump per semver.
> 

With semver, when major is `0`, it is used to increase minor on breaking changes and micro for all other changes. I will update.

hamdikh:
Hey @loganmzz, this looks good overall. Could you add the capabilities gating before we merge?
It is just a guard so the option does not fail at apply time on clusters that do not expose storage.k8s.io/v1/VolumeAttributesClass. We already use .Capabilities elsewhere in the chart, for autoscaling for example, and these are default cluster capabilities so it shouldn't be a problem for helm template.
Let me know once it is updated and I'll re-review.

loganmzz:
Updates:
* Fail if provided but not supported
* Documentation as been aligned for all `volumeAttributesClassName` values fields
* New tests added

hamdikh:
@loganmzz it look good on my end, just a quick rebase on the latest updates + a chart version bump for this to be ready to merge

loganmzz:
> @loganmzz it look good on my end, just a quick rebase on the latest updates + a chart version bump for this to be ready to merge

Done 👍

hamdikh (approved):
LGTM 👍🏽

PR #170, opened 2026-06-01
feat(thanos): add support for image registry

loganmzz:

Add support for custom Docker image repository (e.g. mirror), a standard pattern for "modern" Helm chart.

N/A

N/A

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.21.0
+version: 0.22.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.18.4](https://img.shields.io/badge/Version-0.18.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
+![Version: 0.22.0](https://img.shields.io/badge/Version-0.22.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.41.0](https://img.shields.io/badge/AppVersion-v0.41.0-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
@@ -45,8 +45,8 @@ Kubernetes: `>= 1.30.0-0`
 
 | Repository | Name | Version |
 |------------|------|---------|
-| https://charts.rustfs.com/ | rustfs | 0.3.0 |
-| https://prometheus-community.github.io/helm-charts | kube-prometheus-stack(kube-prometheus-stack) | 85.2.0 |
+| https://charts.rustfs.com/ | rustfs | 0.8.0 |
+| https://prometheus-community.github.io/helm-charts | kube-prometheus-stack(kube-prometheus-stack) | 87.0.1 |
 
 ## Component Overview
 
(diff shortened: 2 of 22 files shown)


loganmzz:
I miss description of breaking changes... But I don't see where to put them 🤔

hamdikh:
Thanks @loganmzz, this is a good direction. Splitting `registry` from `repository` via a shared `thanos.image` helper is the pattern we want. A few things to sort before merge.

## Please rebase first

Could you rebase on the latest `master`? There have been changes since you opened this and I want the diff clean before final review.

## Design decision

I am happy with the explicit split. To keep it unambiguous I want both `registry` and `repository` always set explicitly, no implicit empty registry:

- `global.image`: `registry: quay.io` + `repository: thanos/thanos`, both required. Already how you have it, good.
- The ruler sidecar image should follow the same rule. Right now it is `registry: ""` + `repository: alpine/kubectl`. Please make the registry explicit, e.g. `registry: docker.io` + `repository: alpine/kubectl`, and keep both required in the schema like the global image.

So keep `registry` in the `required` arrays in `values.schema.json` for both blocks. Explicit everywhere is the goal.

## Fold the tag into the helper

While you are in there, please move tag handling into `thanos.image` too so callers do not assemble the reference by hand. And the tag should default to the chart `appVersion` when not set, so the version has a single source of truth in `Chart.yaml` rather than being duplicated in `values.yaml`. That means the helper needs the root context passed in alongside the image dict. Suggested shape:

```gotemplate
{{- /*
thanos.image renders a fully qualified image reference.
Expects a dict with:
  image:   an image dict with registry, repository and optional tag
  context: the root context, used to default the tag to .Chart.AppVersion
*/ -}}
{{- define "thanos.image" -}}
{{- $image := .image -}}
{{- $registry := "" -}}
{{- with $image.registry -}}
{{- $registry = printf "%s/" (trimSuffix "/" .) -}}
{{- end -}}
{{- $tag := $image.tag | default .context.Chart.AppVersion -}}
{{- printf "%s%s:%s" $registry $image.repository $tag -}}
{{- end -}}
```

Call sites then become:

```yaml
image: {{ include "thanos.image" (dict "image" .Values.global.image "context" $) | quote }}
```

and for the sidecar:

```yaml
image: {{ include "thanos.image" (dict "image" .Values.ruler.autoImportPrometheusRules.sidecar.image "context" $) | quote }}
```

With this, `global.image.tag` can become optional in `values.yaml` and the schema (defaults to `appVersion`). The sidecar keeps its own explicit tag, so it never falls back to `appVersion`, which is what we want.

Two things this also fixes versus the current helper: the empty no-op comment block is replaced with a real description, and the whitespace control is corrected. The current `{{ printf ... }}` line has no trimming while the preceding `{{- end }}` leaves a trailing newline, so `include` returns a leading newline and, since callers pipe through `| quote`, you get `image: "\nthanos/thanos:v0.41.0"`. Worth double checking CI on the empty-registry path, that is where it surfaces.

## Breaking change: needs an upgrade note

This changes the values contract. Users who currently pin the full path:

```yaml
global:
  image:
    repository: quay.io/thanos/thanos
```

will now render `quay.io/quay.io/thanos/thanos` since `registry` defaults to `quay.io`. Expected given the split, but it is breaking, so please:

- Treat it as breaking in the changelog / release notes.
- Add a short migration note in the README: registry now lives in `global.image.registry`, repository must be the path without the registry host, and tag defaults to the chart `appVersion`.

## Tests

The updated tests cover the registry-set path, good. Please make sure both the global image and the sidecar assert the full rendered reference (`quay.io/thanos/thanos:<tag>` and `docker.io/alpine/kubectl:<tag>`). Also add a case that leaves `global.image.tag` unset and asserts it resolves to the chart `appVersion`, since that is the new default behaviour.

Thanks again, happy to re-review once it is rebased and these are in.

loganmzz:
@hamdikh 

> Treat it as breaking in the changelog / release notes.

As I stated above, I don't know where to put the breaking change 🤔

loganmzz:
I think I handled all first points:

* I rebased upon `master`
* I made all registry explicit
* I also applied the pattern to `rustfs.initBucket.image`
* I updated tests to check `registry`
* I defaulted image tag to Chart `appVersion`
* I fixed the generated undesired line feeds
* I added an alert in `Upgrading` section
* I also updated `global.image` example in README

hamdikh:
@loganmzz thanks for the contributions 😄

Quick note on the rustfs values section update: rustfs is a dependency used only for CI and tests, and it looks like they don't yet support the added fields (https://github.com/rustfs/rustfs/blob/main/helm/rustfs/values.yaml#L9), just revert this part and we'll be good to go 🥇.

As for the breaking change, the documentation you added covers it well.

loganmzz:
Ok, thought `initBucket` was specific to Thanos chart. Everything related to this should have been reverted!

hamdikh (approved):
LGTM 👍🏽

PR #172, opened 2026-06-12
feat(thanos): add `global.storageClass` for PVCs

Alja9:

Adds `global.storageClass` as a chart-wide fallback for every PersistentVolumeClaim the chart creates: Compactor, Receive (the Ingester in split mode), Ruler and Store Gateway.

A component's own `persistence.storageClass` still wins. When both are empty nothing is rendered, so an install that never set either keeps using the cluster default StorageClass — the rendered output for a default install is byte-for-byte unchanged.

- The fallback lives in a `thanos.storageClass` helper rather than being repeated in four templates. It takes the same `(dict "root" . "key" <component> "persistence" <block>)` shape as the other PVC helpers and absorbs the `thanos.volumeAttributesClassName` include it used to sit next to, so each `volumeClaimTemplate` renders its storage fields from a single include.
- **Scope sweep:** `volumeClaimTemplates` only exist in `compactor/`, `receive/`, `ruler/` and `storegateway/statefulset.yaml`; nothing else in the chart creates a PVC. Query, Query Frontend and BucketWeb are Deployments with no persistence block — BucketWeb reads blocks straight from the object store on each request — so there is no PVC for the fallback to reach. All four PVC-producing components are covered.
- Tests: `global_test.yaml` runs the unset / global-fallback / component-override matrix against all four components at once; each component suite adds its own fallback and override case; Receive covers split mode, where persistence resolves from `receive.ingester.persistence`; Store Gateway covers real sharded rendering via `sharded.hashPartitioning.shards` and asserts every shard picks the class up. `volumeAttributesClassName` is asserted alongside a global `storageClass` since the same helper now renders both.
- Interaction with #164: `global.storageClass` is ignored when a component mounts an `existingClaim`, and a sharded Store Gateway that sets `existingClaim` still resolves the class on the `volumeClaimTemplates` it falls back to. Both are pinned down by tests.
- Docs: the upgrade note and a `Choosing a StorageClass` subsection live in `README.md.gotmpl`, which is what helm-docs generates `README.md` from, so they survive the next `helm-docs` run.

#### Upgrade note

`global` values propagate into subcharts. An umbrella chart that already sets `global.storageClass` for another dependency will now have Thanos render a `storageClassName` on StatefulSets that previously had none, and `volumeClaimTemplates` are immutable, so the upgrade is rejected with `field is immutable`. The README documents how to check for this and the two ways out (pin `persistence.storageClass: ""` per component, or recreate the StatefulSets with `--cascade=orphan`).

#### Checklist

### DIFF
--- a/charts/thanos/templates/_helpers.tpl
+++ b/charts/thanos/templates/_helpers.tpl
@@ -293,6 +293,22 @@ volumeAttributesClassName: {{ . | quote }}
 {{- end }}
 
 {{- /*
+Render the storage-class fields of a PVC spec: `storageClassName`, taken from the
+component's own `persistence.storageClass` and falling back to the chart-wide
+`global.storageClass` when it is empty, followed by `volumeAttributesClassName`.
+Renders nothing when neither is set, so the cluster default StorageClass applies.
+Usage:
+  {{ include "thanos.storageClass" (dict "root" . "key" "compactor" "persistence" .Values.compactor.persistence) }}
+*/ -}}
+{{- define "thanos.storageClass" }}
+{{- $sc := .persistence.storageClass | default .root.Values.global.storageClass }}
+{{- with $sc }}
+storageClassName: {{ . | quote }}
+{{- end }}
+{{- include "thanos.volumeAttributesClassName" (dict "root" .root "key" .key "persistence" .persistence) }}
+{{- end }}
+
+{{- /*
 Render extra volume entries (concatenated) from global, optional parent, then component.
 Usage:
   {{ include "thanos.extraVolumeItems" (dict "root" . "key" "compactor") }}
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.18.0
+version: 0.19.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.37.0
+version: 0.38.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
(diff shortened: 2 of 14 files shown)


hamdikh:
@Alja9 could you rebase on the latest changes i'll take a look asap 👀

Alja9:
Hi @hamdikh ,
it's done, thank you!

hamdikh:
@Alja9 Thanks for this, the direction is right and it lines up with the existing global.* fallbacks in the chart. Two things before merge:

The same five-line block is now repeated in four templates. The chart already has thanos.volumeAttributesClassName for exactly this kind of per-PVC field, so this would be better as a small helper.
All four new tests target compactor only. Please add coverage for receive, ruler and storegateway, storegateway especially since it renders in a shard context with $ scoping.

One more thing: the change only covers compactor, receive, ruler and storegateway. Can you sweep the rest of the chart and apply the same global.storageClass fallback everywhere a PVC can be created, bucketweb included, so the behaviour is consistent across every component rather than only the four stateful ones? If a component deliberately stays out of scope, please say why in the PR description.

Also worth noting in the README upgrade notes: global propagates into subcharts, so anyone already setting global.storageClass in an umbrella chart will now get a storageClassName rendered on existing StatefulSets, and since volumeClaimTemplates are immutable the upgrade will be rejected by the API server. Minor bump is correct, but users should be warned.

Alja9:
Hi @hamdikh , thanks again for the feedback. 
I've updated the PR to address the requested changes based on your reviews. And also updated the PR description.

Could you please take another look when you have a chance? 
I'm not completely sure if my interpretation of the requested scope and implementation matches what you had in mind. Thanks!

charts/thanos/templates/_helpers.tpl
```diff
@@ -261,6 +261,19 @@ volumeAttributesClassName: {{ . | quote }}
...
 
+{{- /*
+Render PVC storage configuration: storageClassName with global fallback and volumeAttributesClassName.
+Usage:
+  {{ include "thanos.storageClass" (dict "root" . "key" "compactor" "storageClass" .Values.compactor.persistence.storageClass "globalStorageClass" .Values.global.storageClass "persistence" .Values.compactor.persistence) }}
+*/ -}}
+{{- define "thanos.storageClass" }}
+{{- $sc := default .globalStorageClass .storageClass }}
+{{- if $sc }}
+        storageClassName: {{ $sc | quote }}
```
hamdikh: instad of having indentation done in the _helper_ like this maybe make use of the `nindent `or `indent `function as it's safer

hamdikh (commented):
let's also add the storageClass change to all of the components using storage :)

charts/thanos/templates/_helpers.tpl
Alja9: sorry, I forgot to notify. I've done revised. please review again. thanks!

hamdikh: don't hesitate to ping me once you've added the remaining components support, if you have any question feel free to reach out on the CNCF slack :) 

Thanks again for your contributions 🔥

hamdikh (approved):

PR #179, opened 2026-06-25
fix(thanos): add ruler replica label for deduplication

### LINKED ISSUE #137: feat: Add automatic replica label support for HA Thanos Ruler
## Problem

When running Thanos Ruler with multiple replicas, the chart currently requires users to manually configure a replica label so that ruler-generated series and alerts can be deduplicated correctly.

For example, users need to add configuration similar to:

```yaml
ruler:
  enabled: true
  extraArgs:
    - --label=thanos_ruler_replica="$(POD_NAME)"
    - --alert.label-drop=thanos_ruler_replica
  extraEnv:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.name
```

Without this, multiple Ruler replicas may produce duplicate rule evaluation series and duplicate alerts unless users know to configure the replica label manually.

The chart already handles a similar pattern for Receive by automatically adding:

```yaml
--label=receive_replica="$(POD_NAME)"
```

and `query.replicaLabels` already includes `receive_replica` by default. Ruler would benefit from the same first-class handling.

## Proposed solution

Add built-in support for a Ruler replica label, for example `thanos_ruler_replica`, that can be used when `ruler.replicaCount > 1`.

The chart could automatically:

1. Add `POD_NAME` to the Ruler container environment.
2. Add `--label=thanos_ruler_replica="$(POD_NAME)"` to the `thanos rule` arguments.
3. Add `--alert.label-drop=thanos_ruler_replica` so Alertmanager can deduplicate alerts from HA Ruler replicas.
4. Add `thanos_ruler_replica` to `query.replicaLabels` so query-time deduplication works for ruler-generated series.
5. Optionally expose a value such as:

```yaml
ruler:
  replicaLabel:
    enabled: true
    name: thanos_ruler_replica
    alertLabelDrop: true
```

For compactor-side deduplication, the chart could also expose a documented opt-in value that adds:

```yaml
--deduplication.replica-label=thanos_ruler_replica
```

to `thanos compact`.

Because Thanos documents compactor deduplication as experimental and irreversible, it may be better for this part to be explicitly opt-in rather than enabled automatically.

## Alternatives considered

The current workaround is to configure `ruler.extraArgs` and `ruler.extraEnv` manually. This works, but it is easy to miss and creates inconsistent behavior compared to Receive, where the chart already provides the replica label automatically.

Users can also manually add the same label to `query.replicaLabels` and `compactor.extraArgs`, but that requires knowing all deduplication layers that need to be configured.

## Additional context

Thanos Ruler HA documentation recommends configuring external labels that identify the HA group and the replica, and dropping the replica label before sending alerts to Alertmanager so alerts can be deduplicated.

Thanos Compactor also supports offline deduplication through `--deduplication.replica-label`, but the documentation marks this as experimental and notes that the process is irreversible. Because of that, query-time deduplication and alert label dropping seem safe to automate, while compactor deduplication should likely be an explicit chart option.

r0bj:

Adds `ruler_replica` support for Thanos Ruler deduplication.

Changes:
- adds `ruler_replica` to default `query.replicaLabels`
- configures Ruler with `--label=ruler_replica="$(POD_NAME)"`
- drops the replica label from outgoing alerts with `--alert.label-drop=ruler_replica`
- updates unit test coverage and README values docs

- fixes #137 

#### Checklist

### DIFF
--- a/charts/thanos/templates/_helpers.tpl
+++ b/charts/thanos/templates/_helpers.tpl
@@ -498,6 +498,15 @@ env:
 {{- end }}
 {{- end }}
 
+{{- define "thanos.podNameEnvBlock" -}}
+env:
+  - name: POD_NAME
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.name
+  {{- include "thanos.extraEnvItems" . | nindent 2 }}
+{{- end }}
+
 {{- define "thanos.extraEnvFromItems" -}}
 {{- $parent := .parent | default "" -}}
 {{- $par := dict -}}
--- a/charts/thanos/templates/_helpers.tpl
+++ b/charts/thanos/templates/_helpers.tpl
@@ -498,13 +498,14 @@ env:
 {{- end }}
 {{- end }}
 
-{{- define "thanos.podNameEnvBlock" -}}
+{{- define "thanos.podEnvBlock" -}}
 env:
   - name: POD_NAME
     valueFrom:
       fieldRef:
         fieldPath: metadata.name
   {{- include "thanos.extraEnvItems" . | nindent 2 }}
+{{- include "thanos.extraEnvFromBlock" . }}
 {{- end }}
 
 {{- define "thanos.extraEnvFromItems" -}}
--- a/charts/thanos/templates/_helpers.tpl
+++ b/charts/thanos/templates/_helpers.tpl
@@ -504,7 +504,7 @@ env:
     valueFrom:
       fieldRef:
         fieldPath: metadata.name
-  {{- include "thanos.extraEnvItems" . | nindent 2 }}
+{{- include "thanos.extraEnvItems" . | nindent 2 }}
 {{- include "thanos.extraEnvFromBlock" . }}
 {{- end }}
 
--- a/charts/thanos/templates/ruler/statefulset.yaml
+++ b/charts/thanos/templates/ruler/statefulset.yaml
@@ -50,6 +50,8 @@ spec:
             - --data-dir=/var/thanos/ruler
             - --objstore.config-file=/etc/thanos/objstore.yml
             - --rule-file=/etc/thanos/rules/*.yaml
+            - --label=ruler_replica="$(POD_NAME)"
+            - --alert.label-drop=ruler_replica
           {{- if .Values.ruler.autoImportPrometheusRules.enabled }}
             - --rule-file=/etc/thanos/rules/generated/*.yaml
           {{- end }}
--- a/charts/thanos/templates/ruler/statefulset.yaml
+++ b/charts/thanos/templates/ruler/statefulset.yaml
@@ -63,8 +63,7 @@ spec:
           {{- range .Values.ruler.extraArgs }}
             - {{ . | quote }}
           {{- end }}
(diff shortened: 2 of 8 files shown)


hamdikh:
@r0bj  let's rebase on this, you're also missing the POD_NAME injection in the env variables for the label, I don't see any issue integrating this, but let's keep the PR updated;

Thanks a lot for your contribution :)

r0bj:
@hamdikh Thanks for the info. I've addressed your comment.

charts/thanos/templates/ruler/statefulset.yaml
```diff
@@ -61,7 +63,12 @@ spec:
           {{- range .Values.ruler.extraArgs }}
             - {{ . | quote }}
           {{- end }}
-          {{- include "thanos.extraEnvBlock" (dict "Values" .Values "component" "ruler") | nindent 10 }}
+          env:
```
hamdikh: @r0bj I would manage this part using a helper as it can get messy, this will simplifiy a bit the readability of the yaml file and permits a good construction of the extra blocks

r0bj: I think I've addressed it in https://github.com/thanos-community/helm-charts/pull/179/changes/fa19dfae21c39992dce4267ee341bd018f795722

charts/thanos/templates/ruler/statefulset.yaml
```diff
@@ -61,7 +63,7 @@ spec:
           {{- range .Values.ruler.extraArgs }}
             - {{ . | quote }}
           {{- end }}
-          {{- include "thanos.extraEnvBlock" (dict "Values" .Values "component" "ruler") | nindent 10 }}
+          {{- include "thanos.podNameEnvBlock" (dict "Values" .Values "component" "ruler") | nindent 10 }}
```
hamdikh: @r0bj let's name it podEnvBlock for clarity, I also don't see the extraEnvBlock anymore, what I suggest is to have the podEnvBlock builds the whole pod injection + extraEnvBlock + extraEnvFrom Block and add those tests to have proper rendering

r0bj: Thanks, I've addressed it in https://github.com/thanos-community/helm-charts/pull/179/changes/ba1c1ca3f416e1e20a9c1854cb7f4c8590b77b32.

charts/thanos/templates/_helpers.tpl
```diff
@@ -498,6 +498,16 @@ env:
 {{- end }}
 {{- end }}
 
+{{- define "thanos.podEnvBlock" -}}
+env:
+  - name: POD_NAME
+    valueFrom:
+      fieldRef:
+        fieldPath: metadata.name
+  {{- include "thanos.extraEnvItems" . | nindent 2 }}
```
hamdikh: let's have a similar indentation for all env blocks and I think that we"ll be good for this one

r0bj: https://github.com/thanos-community/helm-charts/pull/179/changes/7f5233cc549ef07b30bc0bb88c08f09e684f0349

hamdikh (approved):
LGTM 👍🏽

PR #180, opened 2026-06-26
feat(thanos): Per component service accounts

atsai1220:
- Introduce per-component Kubernetes service account so cloud identity can be properly scoped to each component.

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.23.1
+version: 0.24.0
 appVersion: "v0.41.0"
 keywords:
   - thanos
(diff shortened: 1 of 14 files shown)


hamdikh:
Hey @atsai1220, thanks a lot for putting this together. Scoping identity per component is exactly the right direction for this chart, and the least-privilege win on the ruler rule-importer binding is a real improvement. This is a really relevant change.

A few things I would like before we merge:

1. Could you rebase on master? It has moved since you opened this.

2. Split the ServiceAccounts into one file per component (bucket/serviceaccount.yaml, compactor/serviceaccount.yaml, and so on) instead of the single serviceaccount.yaml holding everything. It matches the existing per-component layout and is much easier to maintain.

3. Add a per-component serviceAccount section in values (create, name, annotations) so each component can carry its own cloud identity annotation. Right now every SA gets the same global.serviceAccount.annotations, which means IRSA and GKE Workload Identity would bind the same role to every component, and the per-component goal is lost.

4. Add a clean toggle to disable the global service account creation, decoupled from rbac.create. Today the whole rendering is gated on serviceAccount.create AND rbac.create, so create=true with rbac.create=false leaves pods pointing at SAs that are never created.

5. Add the tests to guard all of the above: each per-component file renders only when its component is enabled, the shared-name path renders a single SA that every pod references, per-component annotations override the global set, and nothing renders when creation is disabled.

Happy to help if any of this is unclear. Thanks again, this is a valuable contribution.

hamdikh:
Hi @atsai1220, thanks for raising this and for the work you put in here.

Per component ServiceAccounts have landed on master with #195. The implementation differs from this branch: per component serviceAccount config (create, name, annotations) with fallback to the global settings, one file per component, a shared ServiceAccount for the store gateway shards, plus values schema updates and unit tests covering the rendering paths. The review points from this PR are all addressed there.

Closing this one since the feature is now in master. Thanks for pushing the topic, it is a good addition to the chart. If you spot anything from your branch that is missing, in particular around receive in split mode, an issue or a follow up PR is welcome.

PR #188, opened 2026-07-17
feat(thanos): allow disabling automountServiceAccountToken

tibuntu:

Adds a new value `global.serviceAccount.automountServiceAccountToken` (default `true`) that is rendered on the ServiceAccount. Since no component sets the field at pod level, all pods inherit it, so setting it to `false` disables mounting the ServiceAccount token chart-wide — useful for hardening deployments that don't need Kubernetes API access.

The default keeps current behavior unchanged (Kubernetes defaults the field to `true`).

Note: with the token unmounted, the Ruler's rule auto-import (which talks to the API server) will not work, so users enabling this need that feature disabled.

This might be conflicting with: https://github.com/thanos-community/helm-charts/pull/180

Unit tests added for both the default and the disabled case; `helm unittest` passes (668 tests). README regenerated with helm-docs v1.14.2 and `values.schema.json` updated.

#### Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.30.0
+version: 0.31.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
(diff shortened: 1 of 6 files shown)


tibuntu:
@hamdikh could you review again?

hamdikh (approved):
LGTM 👍🏽

hamdikh:
looks good @tibuntu  thanks for the contrib 

I'll look into the #180 as it's a nice feature to have for people using workload identity for eg

PR #199, opened 2026-08-15
feat(gateway-api): support timeouts on HTTPRoute rules

### LINKED ISSUE #194: HTTPRoute missing timeouts (HTTPRouteTimeouts)
## Problem

It's not easy to enable `timeouts` for HTTPRoute. The default timeout for us was too short, resulting in `504 upstream timeout` errors from Grafana, so we had to manually deploy an HTTPRoute to fix this, but it would nice to use the helm chart. `HTTPRouteTimeouts` seems to be the only thing non-experimental that is missing from the template. It's isn't Core, but it is Extended (like `name` is), so I don't see any reason not to include it.

- helper template: https://github.com/thanos-community/helm-charts/blob/7ab74f6223fb1e4cb91d308e0891f5f3eb1dbc75/charts/thanos/templates/_helpers.tpl#L870-L880
- Spec: https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httprouterule
  - https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httproutetimeouts

## Proposed solution

- Add `timeouts` to the helper template  (I can prepare and MR for this if that sounds ok)
- OR allow arbitrary keys to accommodate future core, extended, or experimental features.
- OR allow disabling the primary rule and only render the `extraRules`

## Alternatives considered

Use the `extraRules` to configure another rule with the timeout, but the original rule would still be templated and would need to be created as a "dummy" configuration, which isn't ideal.

hamdikh:

`HTTPRouteTimeouts` was the only non-experimental field of `HTTPRouteRule` the chart did not render, so users needing a longer request timeout had to deploy an HTTPRoute outside the chart or add a dummy `extraRules` entry. The reporter hit `504 upstream timeout` from Grafana because of it.

Adds `<component>.httpRoute.timeouts`, rendered on the generated rule for all eight HTTPRoutes (bucketweb, compactor, query, query-frontend, receive, receive router, ruler, storegateway).

```yaml
query:
  httpRoute:
    enabled: true
    timeouts:
      request: 60s
      backendRequest: 55s
```

- fixes #194

- Takes option 1 from the issue (typed `timeouts`) rather than arbitrary keys or a disable-the-primary-rule switch, so the value stays schema-validated like its siblings.
- **GRPCRoute is deliberately left alone**: `GRPCRouteRule.timeouts` is still an experimental field, whereas `HTTPRouteTimeouts` is Extended and stable — same tier as `name`, which the chart already renders.
- Purely additive: `timeouts` defaults to `{}` and nothing is emitted unless set.
- Tests: 3 new cases in `gateway_api_test.yaml` (query timeouts, absent by default, query-frontend `request: 0s`). Full suite `helm unittest charts/thanos` → 746/746.
- The README diff includes pre-existing drift on master: the version badge was stale at `0.32.0` and the kube-prometheus-stack row at `88.0.1`. `helm-docs` corrects both.
- @briend offered to send this in the issue — happy to close this in favour of their PR if they prefer.

#### Checklist

- [x] Chart Version bumped (0.33.0 → 0.34.0)

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.33.1
+version: 0.34.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.33.1](https://img.shields.io/badge/Version-0.33.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
+![Version: 0.34.0](https://img.shields.io/badge/Version-0.34.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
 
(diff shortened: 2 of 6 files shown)


briend:
Thank you @hamdikh !

PR #202, opened 2026-08-18
feat(thanos): make --http.config usable — probe transport/scheme and ServiceMonitor auth

ErvalhouS:
## What

`--http.config` ([docs](https://thanos.io/tip/operating/https.md/)) lets Thanos serve TLS and basic auth on its HTTP server, but the chart can't currently be configured to survive it. This adds the three values needed, changing nothing by default.

## Why it doesn't work today

Thanos passes its whole HTTP mux to exporter-toolkit's `ListenAndServe`, and `/-/healthy` and `/-/ready` are registered on that same mux ([`pkg/server/http/http.go`](https://github.com/thanos-io/thanos/blob/main/pkg/server/http/http.go)). exporter-toolkit's handler has **no path exemptions** — with any `basic_auth_users` entry, every request needs credentials, probes included; and `tls_server_config` makes an HTTP probe fail on scheme.

The chart renders probes as `httpGet` with no `scheme` and no `httpHeaders`, so enabling either half of `--http.config` leaves every component **permanently unready**. `/metrics` scraping breaks too, because the ServiceMonitors had no way to authenticate.

Mounting the config file itself already works via `extraArgs` + `extraVolumes` — it's everything downstream of it that couldn't be expressed.

## What this adds

| Value | Effect |
|---|---|
| `global.probes.tcpSocket` | Swap HTTP probes for a TCP check, which needs no credentials — the fix for `basic_auth_users` |
| `global.probes.scheme` | Probe scheme, for TLS without basic auth. Emitted only when set |
| `global.serviceMonitor.basicAuth` | Credentials for scraping `/metrics`, which `--http.config` also protects |

Probe values resolve **global → parent → component** using `hasKey` rather than `default`, so a component can set `tcpSocket` back to `false` against a global `true` (`default` treats `false` as empty and would ignore it).

Also renders `scheme` and `tlsConfig` on the ServiceMonitors. Both already existed under `global.serviceMonitor` but no template consumed them, so they silently did nothing.

## Compatibility

**Defaults are unchanged.** Rendering the chart without any of these values produces byte-identical output to `master` — verified by diffing `helm template` against a clean checkout with receive, query and ServiceMonitors enabled.

## Testing

- new `tests/http_config_test.yaml`: both probe transports, the component-level override, ServiceMonitor `scheme`/`tlsConfig`/`basicAuth`, and absence when unset
- full suite passes: **809 tests, 15 suites**
- `values.schema.json` updated; README regenerated with `helm-docs`
- chart `version` bumped 0.36.0 → 0.37.0

## Example

```yaml
global:
  probes:
    tcpSocket: true          # --http.config sets basic_auth_users
  serviceMonitor:
    enabled: true
    basicAuth:
      username: { name: thanos-http-auth, key: username }
      password: { name: thanos-http-auth, key: password }

receive:
  extraArgs:
    - --http.config=/etc/thanos/http/http-config.yml
  extraVolumes:
    - name: http-config
      secret: { secretName: thanos-http-config }
  extraVolumeMounts:
    - name: http-config
      mountPath: /etc/thanos/http
      readOnly: true
```

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.38.0
+version: 0.39.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -884,8 +884,8 @@ The table below documents all available values. Top-level keys group settings by
 | global.podAnnotations | object | {} | Annotations added to every pod by default. Component-level annotations are merged on top. |
 | global.podSecurityContext | object | {} | Pod-level security context applied to every pod. Component-level values override this. |
 | global.priorityClassName | string | `""` | Priority class name applied to every pod by default. |
-| global.probes | object | `{"scheme":"","tcpSocket":false}` | Probe defaults shared by every component. A component's own `probes` block overrides these. |
-| global.probes.scheme | string | `""` | Scheme for HTTP probes. Set to `HTTPS` when components serve TLS through `--http.config` (https://thanos.io/tip/operating/https.md/). Omitted when empty, so rendered manifests are unchanged unless this is set. |
+| global.probes | object | `{"scheme":"","tcpSocket":false}` | Probe *transport* defaults shared by every component: `scheme` and `tcpSocket` only. Timings and paths are not inherited from here — set those on each component's own `probes.readiness`/`liveness`/`startup`. Any other key here fails schema validation. |
+| global.probes.scheme | string | `""` | Scheme for HTTP probes. Uppercase, as the kubelet requires — unlike its neighbour `serviceMonitor.scheme`, which is lowercase. Set to `HTTPS` when components serve TLS through `--http.config` (https://thanos.io/tip/operating/https.md/). Omitted when empty, so rendered manifests are unchanged unless this is set. |
 | global.probes.tcpSocket | bool | `false` | Use a TCP probe instead of an HTTP one. Required when `--http.config` sets `basic_auth_users`: Thanos protects `/-/healthy` and `/-/ready` too, so HTTP probes 401. |
 | global.rbac.create | bool | `true` | Create RBAC resources (ClusterRole, ClusterRoleBinding) required by components that need cluster-level access (e.g. Ruler auto-import). |
 | global.resources | object | {} | Default resource requests and limits. Override per component as needed. |
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.38.0](https://img.shields.io/badge/Version-0.38.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
+![Version: 0.39.0](https://img.shields.io/badge/Version-0.39.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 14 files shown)


hamdikh (commented):
Hey @ErvalhouS  
The PR is legit, and looks mosetly good, one thing though, to simplify the usage and enhance documentation we have two smaller things on the probe values:

global.probes only actually behaves as a global for tcpSocket and scheme. The resolution in thanos.httpProbes still picks $probes from component → parent only, so anything else set under global.probes readiness, liveness, startup is silently ignored, and additionalProperties: true on the new schema node means a typo there never surfaces either. The description reads "Probe defaults shared by every component", which promises more than it delivers. Either narrow the wording to make clear it's only the transport, or lock the schema node down to just the two keys so anything else fails loudly.

Second, probes.scheme has no validation, and kubelet requires it uppercase (HTTP/HTTPS) while its neighbour serviceMonitor.scheme requires it lowercase (http/https). Two adjacent values with opposite casing rules is an easy thing to get wrong, and a lowercase value here renders and lints fine it only fails when the API server rejects the manifest. An enum in the schema would catch it at template time.

ErvalhouS:
@hamdikh Both addressed in ff828a4.

**1. `global.probes` over-promising** — went the "lock it down" route rather than only rewording, so a typo fails instead of being ignored: the node is now `additionalProperties: false`, and the description says what it actually is ("Probe *transport* defaults shared by every component: `scheme` and `tcpSocket` only. Timings and paths are not inherited from here — set those on each component's own `probes.readiness`/`liveness`/`startup`.").

```console
$ helm template t charts/thanos --set global.probes.readiness.enabled=true
Error: values don't meet the specifications of the schema(s) in the following chart(s):
thanos:
- global.probes: Additional property readiness is not allowed
```

**2. `probes.scheme` casing** — `enum: ["", "HTTP", "HTTPS"]`, and the comment now names the trap explicitly ("Uppercase, as the kubelet requires — unlike its neighbour `serviceMonitor.scheme`, which is lowercase"):

```console
$ helm template t charts/thanos --set global.probes.scheme=https
Error: values don't meet the specifications of the schema(s) in the following chart(s):
thanos:
- global.probes.scheme: global.probes.scheme must be one of the following: "", "HTTP", "HTTPS"
```

Also moved the `probes:` block out from between the two `serviceMonitor` comment lines where it had landed by accident, and updated the README rows. `helm lint --strict` clean, `helm unittest` 809/809.

One thing I deliberately left out — the per-component `probes` schema nodes are `additionalProperties: true` and don't declare `scheme`/`tcpSocket`, so a component-level `compactor.probes.scheme: https` still slips past the enum. Locking that down means declaring both keys on all nine component nodes (and adding the keys to each component block in `values.yaml`). Happy to do it here if you'd rather it be airtight in one PR, otherwise the global path — the documented one — is covered.

hamdikh:
@ErvalhouS Thanks for closing those two both check out (readiness.enabled=true and scheme=https now fail as shown).

On the third point: could you close the component-level gap too? bucket.bucketweb.probes, compactor.probes, query.probes, queryFrontend.probes, receive.probes, ruler.probes, and storegateway.probes are all still additionalProperties: true and only declare liveness/readiness/startup scheme/tcpSocket aren't in their properties at all, even though _helpers.tpl reads them off $comp.probes as one of the three resolution sources.

That means the override path bypasses everything you just locked down on global.probes: compactor.probes.scheme: https (lowercase) passes schema validation today and only fails once the API server rejects the rendered manifest same failure mode the enum was added to catch, just not caught here.

Would rather this be airtight everywhere than just on the documented global path please add scheme (same enum) and tcpSocket to those seven component probes nodes, keeping additionalProperties: true for the rest of each node. Otherwise this is good to go.

ErvalhouS:
Good catch — confirmed and fixed in 4a73088.

You were right that the override path was unvalidated: `thanos.httpProbes` reads `scheme`/`tcpSocket` off all three sources (`_helpers.tpl`, `global -> $par.probes -> $comp.probes`), but only `global.probes` declared them, so `additionalProperties: true` on the component nodes meant the enum never ran there.

All seven now declare `scheme` (same `"" / HTTP / HTTPS` enum) and `tcpSocket` (boolean), with `additionalProperties: true` kept as you asked. Schema-only change — no template, values or docs delta, so rendered output is unchanged.

Verified per node, before/after:

```
                     scheme:https   scheme:HTTPS   tcpSocket:yes
bucket.bucketweb     caught         ok             caught
compactor            caught         ok             caught
query                caught         ok             caught
queryFrontend        caught         ok             caught
receive              caught         ok             caught
ruler                caught         ok             caught
storegateway         caught         ok             caught
```

```
$ helm template t . --set compactor.probes.scheme=https
thanos:
- compactor.probes.scheme: compactor.probes.scheme must be one of the following: "", "HTTP", "HTTPS"
```

Suite still green: 842 tests, 15 suites.

One thing worth naming so "airtight" is not over-claimed: with `additionalProperties: true` these nodes still accept misspelled keys — `compactor.probes.readiness.timeoutSecond: 10` and `compactor.probes.livenes.enabled: false` both pass validation today and are silently ignored. That is not specific to probes, it is the chart-wide convention: 289 object nodes are `additionalProperties: true` and exactly one (`global.probes`, from this PR) is `false`.

Closing those nodes would be a breaking change — a stray key that is a harmless no-op today becomes a hard `helm upgrade` failure, with renders byte-identical but acceptance not — and it only buys anything if the whole subtree is closed (21 child nodes for probes alone). That is a schema policy call for the chart, not a rider on an `--http.config` feature, so I have opened #203 with the numbers and three concrete options instead of widening this PR. Happy to implement whichever option maintainers pick.

ErvalhouS:
@hamdikh I think this should wrap it, this should be good to trigger workflows by now.

hamdikh (approved):
LGTM 👍🏽

PR #207, opened 2026-08-31
feat(receive): allow the hashring ConfigMap to be managed externally

sebastiangaiser:
Adds receive.hashrings.external.{enabled,configMapName,key}. When enabled, the chart renders no hashring ConfigMap of its own; the standalone StatefulSet and the split Router mount the named ConfigMap under its own key at /etc/thanos/hashrings.json, and the checksum/hashrings pod annotation is omitted. autogen and static have no effect while it is set, and an empty configMapName aborts the render.

Adds receive.hashrings.algorithm, replacing the hard-coded --receive.hashrings-algorithm=ketama, and receive.hashrings.refreshInterval, which emits --receive.hashrings-file-refresh-interval when non-empty. Both apply to standalone and split mode.

This is what a hashring controller such as thanos-receive-controller needs: it rewrites the ring as Receive pods come and go, and the chart-rendered ConfigMap would keep overwriting it.

- fixes #

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.41.1
+version: 0.42.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.41.1](https://img.shields.io/badge/Version-0.41.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
+![Version: 0.42.0](https://img.shields.io/badge/Version-0.42.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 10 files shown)


hamdikh:
@sebastiangaiser The overall shape looks right, and the `required` on `configMapName` is the
correct guard.

One behavioural gap: dropping the `checksum/hashrings` annotation means that in
external mode the ring content is invisible to Helm, so if `refreshInterval` is
left empty the Receive pods will only pick up controller rewrites on restart.
That is the exact failure this feature is meant to avoid, and a README note is
easy to miss. Could we either default `refreshInterval` to something sensible
when `external.enabled` is true, or `fail` when it is empty in that mode?

Heads up on a conflict with #208, which rewrites the same region of
_helpers.tpl and touches the same `--receive.hashrings-algorithm=ketama` line
in both receive/statefulset.yaml and receive/split/router-deployment.yaml.
Whichever of the two lands second will need a real rebase there, not just a
version bump.

One combination is untested across the two PRs: external hashring plus
capnproto. The chart renders no ring at all in external mode, so it cannot
inject `capnproto_address`, and the user has to supply it themselves. Worth a
line in the README section here once #208 is in.

sebastiangaiser:
I checked the refresh point against the Thanos source and the premise does not hold, so I have not added a `fail` or a forced default. Receive builds a `ConfigWatcher` over the hashring file (`cmd/thanos/receive.go:601`), and that watcher is fsnotify-based (`pkg/receive/config.go:188-192`) — it re-reads on every event that is not CHMOD or Remove. `--receive.hashrings-file-refresh-interval` defaults to `5m` and its help text calls it "(used as a fallback)"; the ticker branch in `ConfigWatcher.Run` carries the comment "Setting a new watch after an update might fail. Make sure we don't lose those files forever.", which is precisely the kubelet symlink swap that a ConfigMap update performs. So with `refreshInterval` left empty, a controller rewrite lands via fsnotify, and within 5 minutes at the very worst — not on restart.

The one way the concern would be real is a `subPath` mount, where the kubelet never updates the projected file. This PR does not use one: the standalone StatefulSet mounts the ring through a projected volume with `items`, and the Router through a `configMap` volume with `items`. Both are directory mounts, so updates propagate.

That said, the misleading expectation came from my own README wording, which said to set `refreshInterval` "so Receive picks up the controller's rewrites rather than depending on a restart". That was wrong, and it is what I would have concluded from reading it too. Fixed in the latest push — the bullet now describes the fsnotify watch and the 5m fallback, and says `refreshInterval` only tightens that bound. If you would still prefer a default in external mode on top of that, I am happy to add one, but it would override the Thanos default rather than fix a gap.

Rebase noted: if #208 lands first I'll do a real rebase of this branch on `_helpers.tpl` and the `--receive.hashrings-algorithm` line, not just a version bump.

hamdikh:
Your reading of the Thanos source is right and mine was not. `ConfigWatcher` is fsnotify based with the 5m ticker as the fallback, and both mounts here are directory mounts (projected with `items` on the ingester, configMap with `items` on the router), no `subPath`, so kubelet updates propagate. No forced default and no fail. The reworded README bullet is accurate.

The rest of the change reads well: helpers instead of duplicated conditionals, `required` on `configMapName`, the ConfigMap suppressed so Helm and the controller cannot fight over the same object, and values, schema, README and README.gotmpl all updated with the version bumped.

Two things I would like changed before this goes in:

1. Split mode placement is undocumented. The README example shows `receive.hashrings.external`, while your tests set `receive.ingester.hashrings.external`. Both the router Deployment and the ConfigMap guard resolve through `thanos.receive.cfg`, so anyone who puts the block under `receive.router.hashrings` gets a silent no-op. Could you add a line to the new README section, and ideally a `fail` when it is set under `router`?

2. `omits the hashring refresh interval by default` asserts `notContains` on `--receive.hashrings-file-refresh-interval=5m`, a string the chart never emits in any configuration, so it passes vacuously. It should assert the flag is absent regardless of its value.

Follow-ups, happy to take them here or later:

- A test for the documented precedence: `external.enabled: true` together with `static` or `autogen.enabled`, asserting no ConfigMap is rendered. The README says they are ignored, let's lock that in.
- `install-chart` only exercises the default path. A `ci/` values file that enables external mode against a pre-created ConfigMap would give the feature real install coverage, if you think it is worth the extra job time.
- Still pending from my first pass: the capnproto note in the README once #208 is in.
- Merge order with #208 stands, both touch the same region of `_helpers.tpl` and the same `--receive.hashrings-algorithm=ketama` line in both files, and both target 0.42.0. Whichever lands second needs a real rebase plus a re-bump.

Address 1 and 2 and I am happy to approve.

sebastiangaiser:
Both addressed in the latest push.

1. Placement. `thanos.receive.cfg` now fails with `receive.router.hashrings is not read: configure the hashring under receive.ingester.hashrings in split mode, or receive.hashrings in standalone mode`. The guard sits there because every hashring path resolves through it, and it is gated on `receive.enabled` so a disabled Receive does not fail on an unused value. The README section gained a line explaining why the block belongs under `receive.ingester` in split mode: both workloads resolve the ring from the Ingester config, the Router to route and the Ingester to find its own place in it. There is a test for the rejection.

2. You were right that the assertion was vacuous, and it took a few attempts to find one that is not. `notMatchRegexRaw` reports green even when the flag is present, and a `lengthEqual` on a filtered path is green with the flag and red without it. What works is `notExists` on `spec.template.spec.containers[0].args[?(@ =~ /hashrings-file-refresh-interval/)]`, which is value-independent. I verified it by mutation: with the default `refreshInterval` set to `90s`, the new assertion fails and the old one still passes.

Also added the precedence test from your follow-ups: `external.enabled` together with `autogen.enabled` and a non-empty `static`, asserting no ConfigMap is rendered.

On the `ci/` values file for `install-chart`: I do not think it can work. External mode means the ConfigMap is created by something outside the release, and `ct install` has no way to pre-create it, so the Receive pod would sit in `ContainerCreating` and the install job would fail. Unless you know a hook I am missing, I would leave install coverage on the default path.

The capnproto README note and the rebase against whichever of #207/#208 lands first are still on my list.

sebastiangaiser:
Amended: I added the README line after all, phrased so it does not depend on #208. Rather than naming Cap'n Proto, it states that an externally managed ring is entirely the owner's responsibility, including any per-endpoint fields beyond the address that the hashring format supports, since the chart contributes nothing to a ring it does not render. That stays accurate once #208 lands.

hamdikh (approved):
LGTM 👍🏽

PR #208, opened 2026-08-31
feat(receive): support Cap'n Proto replication

sebastiangaiser:
Adds receive.replication.protocol (protobuf or capnproto) and receive.service.capnprotoPort. Under capnproto the standalone StatefulSet gets --receive.replication-protocol and --receive.capnproto-address, the split Router gets the protocol flag and the split Ingester the address flag, since only the routing side dials its peers. The default protobuf renders no flags at all.

The port is rendered on the container, on the headless Service and, with global.networkPolicies enabled, in the NetworkPolicy. It is left off the ClusterIP Service, because replication addresses one specific peer through its per-pod DNS name and a load-balanced port would reach a different one.

Auto-generated hashrings emit endpoint objects with an explicit capnproto_address instead of plain strings when the protocol is capnproto, which is what the Thanos docs recommend over address inference. Static hashrings are passed through untouched.

- fixes #

#### Checklist

### DIFF
--- a/charts/thanos/Chart.yaml
+++ b/charts/thanos/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: thanos
 description: A helm chart to setup Thanos on Kubernetes, a highly available Prometheus setup with long term storage capabilities.
 type: application
-version: 0.42.0
+version: 0.43.0
 appVersion: "v0.42.4"
 keywords:
   - thanos
--- a/charts/thanos/README.md
+++ b/charts/thanos/README.md
@@ -1,6 +1,6 @@
 # Thanos Helm Chart
 
-![Version: 0.42.0](https://img.shields.io/badge/Version-0.42.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
+![Version: 0.43.0](https://img.shields.io/badge/Version-0.43.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.42.4](https://img.shields.io/badge/AppVersion-v0.42.4-informational?style=flat-square)
 
 <p align="center"><img src="../../docs/imgs/thanos_logo_full.svg" alt="Thanos Logo" width="300"/></p>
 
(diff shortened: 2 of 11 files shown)


hamdikh:
@sebastiangaiser I checked the Service claim and it holds: the port lands on the headless
Service only, and the comment explaining why is in the right place. Nice.

One thing I would like confirmed against a running cluster rather than the
docs. In split mode the Ingester gets `--receive.capnproto-address` but not
`--receive.replication-protocol`, on the reasoning that only the routing side
dials peers. That reads correct to me for a router plus ingester topology, but
it is the kind of asymmetry that is painful to discover in production. Have you
run this in split mode with replication factor greater than 1?

Same conflict note as on #207: the two PRs overlap in _helpers.tpl around the
hashring helpers and on the `--receive.hashrings-algorithm=ketama` line, which
#207 templates and this one keeps hard coded. Once one merges the other will
need a proper rebase, and ideally a test covering external hashring plus
capnproto together, since the chart cannot inject `capnproto_address` into a
ring it does not render.

sebastiangaiser:
Yes — we run exactly this topology in production: split mode with a router Deployment and an ingester StatefulSet, replication factor 2, and Cap'n Proto replication enabled for one of our larger tenants. It has been running that way for a while, which is what prompted me to upstream it.

The asymmetry matches what we run. Our own manifests pass `--receive.replication-protocol=capnproto` on the routing side only, and the ingesters listen on the Cap'n Proto port without carrying the protocol flag.

One difference worth naming: our manifests also pass `--receive.capnproto-address` on the router. That turns out to be a no-op — the flag defaults to `0.0.0.0:19391` in Thanos (`cmd/thanos/receive.go:1059`), so the Cap'n Proto server binds there whether or not it is passed. The chart omitting it on the router is therefore equivalent to what we run, and I would rather not render a flag that only restates the default.

On the combined test: it needs both features present in one branch, so I will add it when I rebase whichever of the two PRs lands second.

hamdikh:
@sebastiangaiser let's rebase on your latest updates :) i'll take a look once that's done 

Thanks for your contributions ! 🥇

sebastiangaiser:
Rebased onto master now that #207 is in. Chart is at `0.43.0`, and the `_helpers.tpl` and `--receive.hashrings-algorithm` overlaps you flagged are resolved — the hashring helpers from #207 and the Cap'n Proto ones now sit side by side, and thanos.receive.hashrings injects capnproto_address into both ring shapes it renders.

The combined test you asked for is in, since both features finally live in one tree: external ring plus capnproto asserts that no ConfigMap is rendered while the workload is still fully configured for Cap'n Proto and mounts the foreign ConfigMap. The README bullet is now concrete about it too — rings the chart does not render (static and external) get no capnproto_address, so their owner has to supply it. That also closes the follow-up you left on #207.

hamdikh (approved):
LGTM 👍🏽

Issue #173, opened 2026-06-15
Support Thanos Ruler as Deployment


jkroepke:
## Proposed solution

Since Thanos Ruler can work stateless, it would be great if Ruler can be deployed as Deployment as well.

Stateful Sets are sometimes bit clumpy and if there is no need for local storage, I would like to avoid them.

hamdikh:
@saswatamcode @GiedriusS WDYT ? Do you think this would be a good fit for our helm chart ? If so, I can plan to implement this option

PR #209, opened 2026-08-31
feat(ruler): support stateless mode via remote write

sebastiangaiser:
Adds ruler.remoteWrite. When enabled, the Ruler is passed --remote-write.config-file instead of --objstore.config-file and mounts the remote-write Secret in place of the object store one, so evaluation results go to a remote-write endpoint rather than the bucket. --data-dir and the PersistentVolumeClaim stay, since the WAL still needs them.

The config can carry credentials for the write endpoint, so it follows the same contract as global.objstore: createSecret defaults to false and the chart only mounts secretName/secretKey, with an inline config rendered into a Secret when createSecret is true. The pod checksum annotation is rendered for the inline config only, as an externally managed Secret is unknown at render time.

Query no longer registers the Ruler as a --endpoint while it runs stateless, because a stateless Ruler serves no Store API. Both that endpoint and the Ruler objstore mount arrived in #155; neither changes in the default stateful mode.

Related to #173, which asks for the Ruler as a Deployment: stateless mode is the part that makes running it without a StatefulSet meaningful.

- fixes #

#### Checklist


hamdikh:
@sebastiangaiser The stateful path is untouched and both branches are covered by tests, which is
what I wanted to see given this reaches into the Query deployment.

Two things:

`ruler.remoteWrite.secretName` defaults to a non-empty
"thanos-ruler-remote-write". So enabling stateless mode without creating the
Secret first renders a valid manifest and then hangs the pod waiting for a
Secret that does not exist, instead of failing at template time. I know this
mirrors `global.objstore`, so tell me if the consistency is deliberate, but I
would rather have `required` here since the failure mode is a silent stuck pod.

The pod annotations refactor from a literal `annotations:` key to the `$anns`
dict changes rendered output beyond the scope of the PR: with stateless mode,
an external Secret and no `global.podAnnotations`, the `annotations:` key
disappears from the pod template entirely. Harmless to Kubernetes, but it will
show up as an unexplained diff for users. Either keep an empty map or call it
out in the changelog entry.

Minor overlap with #210, which also edits query/deployment.yaml. Different
hunks, but worth re-running the query tests after whichever rebases second.

sebastiangaiser:
Going to resolve merge conflicts when #207/#208 are merged. ⌛

sebastiangaiser:
Rebased onto master now that #207 and #208 are in, chart at 0.44.0. Both of your points are addressed.

`secretName` no longer has a default. A new `thanos.ruler.remoteWriteSecretName` helper resolves it: an explicit name always wins, `createSecret: true` derives `<fullname>-ruler-remote-write`, and with neither the render fails with `ruler.remoteWrite.secretName is required when ruler.remoteWrite.createSecret is false`. You were right that the stuck pod is the worse failure mode, and it is worth noting that `required` alone would not have caught it: the default was non-empty, so it never fired, and whether the Secret actually exists is not knowable at template time. Removing the default is what makes the check bite.

That does diverge from `global.objstore`, which still defaults `secretName` to `thanos-objstore` and has the identical failure mode for every stateful component. I left that alone here rather than widening the PR, but I am happy to send the same treatment for the object store separately if you want the two to line up again.

On the annotations: I took the first of your two options and kept the key, rendering `annotations: {}` when there is nothing to checksum. That keeps the pod template's shape identical to master in every combination, so there is no unexplained diff for users and no changelog entry needed.

Tests: three new ones cover the derived name, an explicit name winning over it, and the missing-name failure. Four existing ones had to start naming the Secret — they were previously riding on the default, which is exactly the behaviour you asked to remove. The suite is at 884, and the default render plus a plain stateful Ruler stay byte-identical to master.

hamdikh:
@sebastiangaiser Thanks, the Secret naming and the annotations shape both look right now `annotations: {}` keeps the pod template identical in every combination, which is what I was after.

One thing left, and it's the Query hunk.

RegisterRulesServer is registered unconditionally in cmd/thanos/rule.go, and info.WithRulesInfoFunc() is always in infoOptions  only the Store server is gated on `tsdbDB != nil`. So a stateless Ruler does still serve the Rules API. On the Query side `--endpoint` is a generic Thanos API endpoint, not a Store endpoint: GetRulesClients() filters on HasRulesAPI() alone, isQueryable() checks health and never HasStoreAPI(), and apisPresent() lists rulesAPI as its own capability. Rules-only endpoints are a supported case.

The premise is correct "no Store API" but the conclusion doesn't follow from it. Dropping the endpoint means Query's /rules, /alerts and the corresponding /api/v1 routes silently stop showing the stateless Ruler's rules and alerts. Keeping it costs nothing: with no StoreInfo advertised, Query never asks it for series. "Query it through the receiver its results land in" covers the samples, but rule and alert state lives in the Ruler.

Could you drop the query/deployment.yaml change and the test that asserts it, and adjust that README bullet? Everything else here is opt-in; this is the one hunk that changes behaviour for the Query UI.

Minor, separate: createSecret: true with `config: ""` renders a Secret with an empty key, and Thanos ignores an empty remote-write config and falls back to its own TSDB with no objstore config that means evaluation results go nowhere, silently. Same class of failure as the stuck pod we just removed. A required on config when createSecret is true would close it.

(Also test-only: ruler_stateless_test.yaml uses 19291 for the Receive remote-write URL, the chart defaults to 10908 the README example has it right.)

sebastiangaiser:
You are right, and I checked every step of it against v0.42.4 before changing anything. `RegisterRulesServer` is in the unconditional `options` slice (`cmd/thanos/rule.go:750`) and `infoOptions` starts as `[]info.ServerOptionFunc{info.WithRulesInfoFunc()}` (`:756`); only the block below it, guarded by `if tsdbDB != nil` (`:757`), adds the label set, `WithStoreInfoFunc` and the Store server. On the Query side `GetRulesClients()` filters on `HasRulesAPI()` alone (`pkg/query/endpointset.go:557-567`), `isQueryable()` is `isStrict || LastError == nil` with no Store check (`:729`), and `apisPresent()` lists `rulesAPI` next to `storeEndpoints` (`:908`). So the premise held and my conclusion did not: a rules-only endpoint is a supported case, and dropping it would have taken the stateless Ruler's rules and alerts out of Query's UI without saying so.

The `query/deployment.yaml` hunk is reverted — that file is now byte-identical to master — and the test that asserted the removal is replaced by one asserting the endpoint stays in stateless mode. The README bullet now says what actually happens: the Ruler remains an endpoint, advertises no Store API so Query never asks it for series, and keeps serving the Rules API so its rules and alerts stay visible.

On the empty config: fixed the same way as the Secret name. `createSecret: true` with an empty `config` now fails the render with `ruler.remoteWrite.config is required when ruler.remoteWrite.createSecret is true`, rather than producing a Secret with an empty key. Both guards live in `thanos.ruler.remoteWriteSecretName`, which every path resolving the Secret goes through.

Test port corrected to 10908 in all four places.

Suite is at 884. The default render and a plain stateful Ruler are byte-identical to master, and the only behavioural change left in this PR is opt-in via `ruler.remoteWrite.enabled`.

hamdikh (approved):
LGTM 👍🏽

PR #210, opened 2026-08-31
feat(query): support file-based endpoint discovery

sebastiangaiser:
Adds query.endpoints.sdConfig. When existingConfigMap or config is set, Query is passed --endpoint.sd-config-file, plus --endpoint.sd-config-reload-interval when reloadInterval is set, and mounts the ConfigMap at /etc/thanos/endpoints. autogen and static endpoints are still rendered as --endpoint arguments alongside it.

existingConfigMap points at a ConfigMap the chart never writes, which is the case that makes file SD worthwhile: the set of Store API servers can change without a helm upgrade. Setting config instead renders <fullname>-query-endpoints from inline definitions in Prometheus file-SD target format.

The ConfigMap is mounted as a directory rather than through subPath, because the kubelet stops updating a subPath-projected file after its first version and Query would keep re-reading a frozen config.

While file SD is enabled the Deployment renders its own volumes and volumeMounts keys and appends the extraVolumes/extraVolumeMounts items to them, since thanos.extraVolumesBlock and thanos.extraMountsBlock render those keys themselves and a second one would be invalid YAML.

- fixes #

#### Checklist


hamdikh:
@sebastiangaiser The subPath reasoning in the description is correct and I am glad it is written
down.

My main concern is the fork in query/deployment.yaml. Because
`thanos.extraVolumesBlock` and `thanos.extraMountsBlock` render the `volumes:`
and `volumeMounts:` keys themselves, this PR ends up with two parallel branches,
one calling the block helpers and one calling the item helpers. That works, but
it means any future change to those helpers has to be mirrored into the fork,
and the next component that needs a chart owned volume will copy the pattern.
#209 is independently working around the same limitation for the Ruler, which
suggests the helpers are the actual problem.

Could you split this: a first PR that makes `extraVolumesBlock` and
`extraMountsBlock` composable (or introduces a helper that takes chart owned
items plus the extras and renders the key once), then this feature on top using
the single path? Happy to discuss the shape before you write it.

Two smaller ones:

- configmap-sd.yaml and deployment.yaml read
  `.Values.query.endpoints.sdConfig.*` directly, while #207 and #208 use `dig`
  with defaults for the same kind of nested access. Worth matching for
  consistency and to survive a user nulling the block.
- Unlike #209 there is no checksum annotation for the inline rendered
  ConfigMap. It works because Thanos reloads the file on its own default
  interval, but two PRs in the same release handling config changes differently
  will confuse people later. Either add it or add a comment saying why it is
  not needed.

sebastiangaiser:
Going to resolve merge conflicts when #207/#208 are merged. ⌛

sebastiangaiser:
Rebased onto master now that #207, #208 and #209 are in, chart at 0.45.0. The two smaller points are done; the helper split I would like to agree on before writing it, see below.

`dig` with defaults is in, in both `configmap-sd.yaml` and `deployment.yaml`, matching how #207 and #208 read nested values. In fairness it buys consistency rather than protection: `sdConfig` and its `key` are in the schema's `required` lists, so nulling either is already rejected with `at '/query/endpoints/sdConfig': missing property 'key'` before a template ever runs.

No checksum annotation, and now with a comment in the template saying why. Query re-reads the file at runtime — `newEndpointConfigProvider` wires the config file into `extkingpin.PathContentReloader` (`cmd/thanos/endpointset.go`), which logs `reloading endpoint config` — so a checksum would restart pods on a change the process is designed to pick up live, and would defeat the point of file SD. The contrast with #209 is real rather than inconsistent: `cmd/thanos/rule.go` uses no reloader at all and reads the remote-write config once at startup, so there a restart is the only way a change takes effect.

On splitting out the helpers: I mapped the call sites first, and the picture argues for going further than composability. Of the eight templates that use these helpers, six already render `volumes:` and `volumeMounts:` themselves and append `extraVolumeItems` / `extraMountItems` — bucket, compactor, receive, the split router, ruler and storegateway. Only `query` is purely block-based, because it is the only component with no chart-owned volume. And `query-frontend` already carries exactly the fork you are objecting to here: `if cacheConfig` renders its own key plus the item helpers, `else` falls back to the block helpers.

So the fork is not a new pattern this PR introduces, it is the third instance. That suggests removing the block helpers entirely rather than making them composable: let Query render its own keys like everyone else, omitted when there is nothing to render, and drop the branch in `query-frontend`. One pattern instead of two, at the cost of a slightly larger refactor PR. If you prefer the narrower option — a helper taking chart-owned items plus the extras and rendering the key once — that works too and I will send it that way. Tell me which shape you want and I will open it, with this PR rebased on top.

hamdikh:
@sebastiangaiser Thanks for the rebase and for mapping the call sites. Before the helper question (see the end), one blocker that I don't think was visible from the templates.

**The config format is wrong.** In v0.42.4, `--endpoint.sd-config-file` doesn't take Prometheus file-SD target groups. Thanos reads the file into its own [`EndpointConfig`](https://github.com/thanos-io/thanos/blob/v0.42.4/cmd/thanos/endpointset.go#L61-L72) struct ([`parse`](https://github.com/thanos-io/thanos/blob/v0.42.4/cmd/thanos/endpointset.go#L172-L182)):

```yaml
endpoints:
  - address: thanos-sidecar-0.prometheus:10901
    # strict, group, service_config, client_config are optional
```

The Querier docs show this format under [Per-Endpoint TLS Configuration](https://thanos.io/v0.42/components/query.md/#per-endpoint-tls-configuration). The `- targets: [...]` format is documented under [File SD](https://thanos.io/v0.42/components/query.md/#file-sd), but that section is about the deprecated `--store.sd-files` flag (see [Flags](https://thanos.io/v0.42/components/query.md/#flags)). The README in this PR links to that same page, which is probably where the two flags got mixed up.

I rendered the README example and ran the [v0.42.4 release binary](https://github.com/thanos-io/thanos/releases/tag/v0.42.4) with it. Query exits at startup:

```
unable to load config file
unable to unmarshal config content: .../endpoints.yml
yaml: unmarshal errors:
  line 1: cannot unmarshal !!seq into main.EndpointConfig
```

The same command with the `endpoints:` format starts normally. So anyone who copies the example from the README or `values.yaml` ends up with a crash-looping Query. CI stays green because the install tests never enable `sdConfig`, and the unit tests only check the rendered strings.

Could you:

- Switch the examples in `README.md.gotmpl`, the `values.yaml` comment, the schema/value descriptions and the test fixtures to the `endpoints:` format. Drop "Prometheus file-SD format" and point the link at [#per-endpoint-tls-configuration](https://thanos.io/v0.42/components/query.md/#per-endpoint-tls-configuration) instead.
- Mention that `default_client_config` and the per-endpoint `client_config` live in this file too. The `--grpc-client-tls-*` flags are [deprecated after v0.43.0](https://thanos.io/v0.42/components/query.md/#flags), and Query already [logs a warning](https://github.com/thanos-io/thanos/blob/v0.42.4/cmd/thanos/query.go#L274) pointing to `default_client_config` in `endpoint.sd-config-file`, so this file is where TLS settings are heading.
- Enable `query.endpoints.sdConfig.config` in `ci/ci-values.yaml` with a valid entry (the in-chart Store Gateway service works), so `ct install` actually starts Query with the flag and would have caught this.

**Things I checked that do hold up:**

- Live reload through the directory mount. [`PathContentReloader`](https://github.com/thanos-io/thanos/blob/v0.42.4/pkg/extkingpin/path_content_reloader.go#L44-L99) re-reads the path on every interval and compares a sha256 of the contents, so it follows the kubelet's `..data` symlink swap. I simulated that swap, and Query logged `reloading endpoint config` and picked up the new endpoint on the next tick. The subPath reasoning is right in practice.
- The case for no checksum annotation. Given that reloader ([wired in here](https://github.com/thanos-io/thanos/blob/v0.42.4/cmd/thanos/endpointset.go#L262)), a pod restart would add nothing. Thanks for the comment in the template.
- Combining with `--endpoint`. [`addStaticEndpoints`](https://github.com/thanos-io/thanos/blob/v0.42.4/cmd/thanos/endpointset.go#L184-L209) merges the flag endpoints into the file's list on startup and on every reload, so autogen and static keep working alongside it. Note that `--endpoint` is now [marked deprecated](https://thanos.io/v0.42/components/query.md/#flags) upstream. That's out of scope here, but worth knowing for a later PR.
- The empty `reloadInterval` falls back to the [`5m` default](https://github.com/thanos-io/thanos/blob/v0.42.4/cmd/thanos/query.go#L194), as the values description says.
- `dig`: agreed, it's consistency more than protection given the schema. Fine as it is.

**On the helpers:** your mapping is accurate, but I read it the other way. The six components that only use the item helpers always have a volume the chart adds itself (objstore, data, and so on), so they never render an empty `volumes:` key and don't need the emptiness check. Query and query-frontend are the only components whose own volumes are optional, and that's the case the block helpers exist for. Removing them would move that check (global, parent and component extras, plus the chart's own condition) into each template. That's more duplication than the fork we have now, and the next component with an optional volume would have to copy it.

So let's go with the narrower option: add an optional `items` argument to `thanos.extraVolumesBlock` and `thanos.extraMountsBlock` holding the chart's own entries. The helper renders the key once when there is anything to render. Both query-frontend and Query then take a single path:

```yaml
{{- $volumes := "" }}
{{- if .Values.queryFrontend.cacheConfig }}
{{- $volumes = printf "- name: cache\n  configMap:\n    name: %s-cache" (include "thanos.compName" (list . "query-frontend")) }}
{{- end }}
{{- include "thanos.extraVolumesBlock" (dict "root" . "key" "queryFrontend" "items" $volumes) | nindent 6 }}
```

I tried this on master with query-frontend. It renders the same manifests with and without `cacheConfig` and with component and global extras, and the existing tests pass untouched. It's a small change (about 16 lines). Please send it as its own PR, converting query-frontend, then rebase this one on top so the Query deployment uses the same single path.

sebastiangaiser:
You are right, and thank you for actually running it — that was a real bug, not a doc nit. I confirmed it in the source before changing anything: `EndpointConfig` has exactly two top-level keys, `default_client_config` and `endpoints`, and `parse()` does a plain `yaml.Unmarshal` into that struct, so a top-level sequence cannot land in it. I had taken the format from the File SD section of the Querier docs without noticing that section belongs to the deprecated `--store.sd-files`, which is exactly the mix-up you guessed.

Fixed in the latest push: the README example, the `values.yaml` example and description, the schema description and three test fixtures now use

```yaml
endpoints:
  - address: thanos-sidecar-0.prometheus:10901
```

The README link points at Per-Endpoint TLS Configuration, the phrase "Prometheus file-SD format" is gone, and a third bullet says gRPC client TLS belongs in this file as `default_client_config` or per-endpoint `client_config`, since the `--grpc-client-tls-*` flags are on their way out.

`ci/ci-values.yaml` now enables `sdConfig.config` against the in-chart Store Gateway. The inline config goes through `tpl`, so it can use the chart's own naming helper:

```yaml
query:
  endpoints:
    sdConfig:
      config: |
        endpoints:
          - address: '{{ include "thanos.compName" (list . "storegateway") }}:10901'
```

That immediately earned its keep: with the ConfigMap actually rendering, `helm lint` caught a second bug of mine in the same template. The `{{- /* ... */ -}}` comment I had added right after the `---` separator was eating the newline, so the document started as `---apiVersion: v1`. Moved above the `if`, and both `helm lint` runs are clean now. Your point about the render path never being exercised was more accurate than either of us knew.

On the helpers: your reading is better than mine. I was looking at which templates call which helper and missed that the six item-only components always have a chart-owned volume, so the emptiness check never applies to them — which is precisely the case the block helpers exist for. Removing them would push that check into every template. I will send the `items` argument as its own PR converting query-frontend, then rebase this one on top so Query takes the single path too.

Also noted for later, outside this PR: `--endpoint` being deprecated upstream means `autogen` and `static` will eventually want to move into this same config file rather than staying flags.

hamdikh (approved):
LGTM 👍🏽

PR #215, opened 2026-09-16
feat(bucket-replicate): add a component for thanos tools bucket replicate

sebastiangaiser:
Adds bucketReplicate, a Deployment running `thanos tools bucket replicate` from global.objstore into a second bucket configured under bucketReplicate.objstoreTo, with the same createSecret contract as the source: the Secret is managed externally unless explicitly opted in. Ships the usual Service, ServiceAccount, ServiceMonitor, NetworkPolicy and PodDisruptionBudget for the component.

Block selection is exposed as typed values: resolutions, compaction.levels, compaction.min, compaction.max, matcher, minTime, maxTime and ignoreMarkedForDeletion. --single-run and --id are deliberately not exposed, because both make the process exit after one pass and a Deployment would restart it forever; those belong in a Job.

The chart has shipped thanos-bucket-replicate alerting rules since #108 while stating that the component is not deployed here. Their toggle now accepts null, which follows bucketReplicate.enabled, so the group appears with the component and an explicit true still alerts on a bucket-replicate running outside the chart.

- fixes #

#### Checklist


hamdikh:
Hi @sebastiangaiser , thanks a lot for this contribution and for the detailed description! 🙏

Before diving into the review, could you help us understand the motivation behind it? The "What this PR does / why we need it" section is empty, and adding a new component means more surface to maintain in the chart, so we'd like to understand the use case first:

1. What is the concrete scenario you are solving? For example cross region DR, migration between providers or buckets, a read replica for another team, or something else?
2. Why run it as a long lived Deployment rather than a Job or CronJob? For a one off migration, a Job with `--single-run` seems like a better fit, so knowing that you need continuous replication would help us weigh the design.
3. Are you running this setup in production today? Any feedback on resource usage or behaviour with large buckets would be useful.
4. Is there a related issue or discussion where other users asked for this?

Once the use case is clear, we can move forward with the review. Thanks again!

PR #216, opened 2026-09-16
feat(helpers): let extraVolumesBlock and extraMountsBlock take the chart's own items

sebastiangaiser:
This is the split #210 asked for. Once it is in, I will rebase #210 on top so the Query deployment uses the same single path instead of the fork it currently carries.

Verification: the default render, query-frontend without volumes and query-frontend with only cacheConfig are byte-identical to master. In the combinations where the block helpers were already in use — global extras, component extras, or both — the raw output loses two to four whitespace-only lines, because the helpers no longer emit a leading newline that nindent turned into an indented blank line. Every difference is a removed whitespace line; the parsed manifests are identical in all six combinations I checked.

thanos.extraVolumesBlock and thanos.extraMountsBlock render the volumes and volumeMounts keys, which forced every component with a conditional volume of its own to fork: one branch writing the key by hand plus the item helpers, another falling back to the block helpers. query-frontend carried that fork for its cache ConfigMap.

Both helpers now take an optional items argument holding the component's own entries as rendered YAML. They are emitted first, followed by global, parent and component extraVolumes, and the key is written once when there is anything to write. query-frontend takes that single path.

The helpers also stop emitting a leading newline, which nindent turned into a whitespace-only line in the rendered output. Parsed manifests are unchanged in every combination of cacheConfig and global or component extras; the raw output loses two to four whitespace-only lines where the block helpers were already in use.

Existing tests pass untouched. Four new ones cover the matrix: neither, own entries only, extras only, and both with their ordering.

- fixes #

#### Checklist


hamdikh (approved):
LGTM 👍🏽

PR #182, opened 2026-07-01
feat: add extraDeploy support

localleon:
#### What this PR does / why we need it:

Adds Bitnami-style extraDeploy support to the Thanos chart so users can render extra Kubernetes manifests alongside the release.
  - Added a dedicated templates/extra-list.yaml renderer for .Values.extraDeploy
  - Documented extraDeploy in values.yaml
  - Updated values.schema.json
  - Regenerated chart README.md
  - Added helm-unittest coverage for empty and templated extra objects

  Validation:

  - helm lint charts/thanos passes
  - schema JSON parses cleanly
  - README regenerated via helm-docs

Unfortunatly ct linting does not work. Seems like the rustFS repo moved? 

```
$ ct lint
Linting charts...
>>> helm version --template {{ .Version }}
>>> git rev-parse --is-inside-work-tree
>>> git rev-parse --verify origin/master
>>> git merge-base origin/master HEAD
>>> git diff --find-renames --name-only c5f5072690773301b7cf3cc8d4df8572011c06d7 -- charts

------------------------------------------------------------------------------------------------------------------------
 Charts to be processed:
------------------------------------------------------------------------------------------------------------------------
 thanos => (version: "0.23.1", path: "charts/thanos")
------------------------------------------------------------------------------------------------------------------------

>>> helm repo add thanos-community https://thanos-community.github.io/helm-charts
"thanos-community" has been added to your repositories
>>> helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
"prometheus-community" already exists with the same configuration, skipping
>>> helm repo add rustfs https://charts.rustfs.com/
Error: looks like "https://charts.rustfs.com/" is not a valid chart repository or cannot be reached: Get "https://charts.rustfs.com/index.yaml": stopped after 10 redirects

------------------------------------------------------------------------------------------------------------------------
No chart changes detected.
------------------------------------------------------------------------------------------------------------------------
Error: failed linting charts: failed adding repo: rustfs=https://charts.rustfs.com/: failed waiting for process: exit status 1
failed linting charts: failed adding repo: rustfs=https://charts.rustfs.com/: failed waiting for process: exit status 1
```

#### Which issue this PR fixes

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*

- fixes #181

#### Checklist

[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]

- [X] [DCO](https://developercertificate.org) signed
[...]

PR #218, opened 2026-09-16
feat: add support for arbitrary extra resources through extraResources

skoef:
Adds extraResources to define arbitrary additional resources to be managed by the chart.

I'm running Thanos with an index/bucket cache configured to use Redis (or actually Valkey) which is deployed together with Thanos. The Thanos chart is managed by ArgoCD, but currently I manage the ConfigMap with the caching configuration separately. I would much rather manage the ConfigMap together with Thanos and making it possible to manage arbitrary resources with the chart adds just this flexibility

I didn't raise an issue for this.

#### Checklist


hamdikh:
Hi @skoef, thanks for the contribution!

I don't really see why managing extra resources through the Thanos Helm chart would help here. The same idea was already discussed in #182 (extraDeploy support), and we decided not to address it as it falls outside the scope of the Thanos Helm chart. The chart should focus on deploying and configuring Thanos components, not act as a generic way to ship arbitrary Kubernetes objects.

For your use case, you can keep the ConfigMap alongside Thanos in ArgoCD without changing the chart, for example with a multi source Application, an umbrella chart wrapping this one, or a small Kustomize overlay.

Closing this one for that reason, but feel free to open an issue if you think something is missing in the chart's native cache configuration.

skoef:
@hamdikh too bad you won't merge such a generic solution. I really like that there is an "official" thanos helm chart now and I would really like to adopt it. As mentioned in #182, it makes customizing the Thanos deployment so much easier.  The fact that I filed the 2nd MR on this topic in a short period of time might speak for itself.

Who knows, some day...

PR #220, opened 2026-09-16
feat(values): Add ability to specify hostUsers at pod-level specifica…

### LINKED ISSUE #219: Add hostUsers flag to pod specifications
## Problem
OpenShift SCC v3 requires [user namespaces](https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/) by disabling hostUsers in pod specification. It increases the security.

## Proposed solution
Add the hostUsers boolean as a Helm variable to values.yaml and implement it in helpers as pod level security context.

## Alternatives considered

Currently on OpenShift we can do two things to solve the problem.
- Decrease security requirements (modify SCC)
- Use a Kyverno policy that patches the resources at the time of deployment (it can cause other kind of problems when using GitOps)

adns44:
Add ability to configure pod-level user namespaces (primarily needed by OpenShift SCC).

fixes #219 [Add hostUsers flag to pod specifications](https://github.com/thanos-community/helm-charts/issues/219)

#### Checklist

[Place an '' (no spaces) in all applicable fields. Please remove unrelated fields.]


adns44:
@hamdikh can you check it pls?

hamdikh:
can you rebase on the latest updates @adns44  ?

adns44:
> can you rebase on the latest updates @adns44 ?

Yes, done. :) @hamdikh

hamdikh:
Thanks for picking this up @adns44, and for the quick rebase! 🙏

I tried the branch locally. Lint, all unit tests and helm-docs pass, and by default nothing new is rendered, so existing installs are unaffected. Global → component precedence works for query, compactor, receive (ingester) and storegateway.

A few changes are needed before we merge:

**1. Router Deployment doesn't render `hostUsers` (blocking)**

`receive.router.hostUsers` was added to `values.yaml` and the README, but `templates/receive/split/router-deployment.yaml` never calls the helper. In split mode the router is the only workload without `hostUsers`, even with `global.hostUsers=false`. Please add this next to the existing `podSC` include:

```yaml
{{- include "thanos.podHostUsers" (dict "root" . "key" "router" "parent" "receive") | nindent 6 }}
```

**2. Unit tests**

Please add a few cases under `charts/thanos/tests/`:
- no `hostUsers` rendered by default
- `global.hostUsers: false` reaches every workload
- a component value overrides global (e.g. `global: false`, `query: true`)
- the router falls back to `receive.hostUsers` in split mode

**3. Schema**

`values.schema.json` doesn't define `hostUsers`, so anything is accepted (e.g. `hostUsers: yes` renders as-is). Please add it as `"type": ["boolean", "null"]` wherever `podSecurityContext` is defined, both global and per component.

**4. README type**

The generated table shows `string` / `nil`. Starting the comments with `# -- (bool) Configure hostUsers ...` makes helm-docs show `bool`.

**Nits (optional)**
- In `thanos.podHostUsers`, `$root.Values.nonExistentKey` is used to get a nil value. `{{- $hu := "" -}}` with a `$found` flag, or checking `hasKey` directly, would be easier to read. The indentation of the `else` is also off.
- Could you shorten the PR title so it isn't cut off? It becomes the squash commit message. For example: `feat: add hostUsers to pod specs`.

With these done it's good to go from my side. Thanks again!

adns44:
> Thanks for picking this up @adns44, and for the quick rebase! 🙏
> 
> I tried the branch locally. Lint, all unit tests and helm-docs pass, and by default nothing new is rendered, so existing installs are unaffected. Global → component precedence works for query, compactor, receive (ingester) and storegateway.
> 
> A few changes are needed before we merge:
> 
> **1. Router Deployment doesn't render `hostUsers` (blocking)**
> 
> `receive.router.hostUsers` was added to `values.yaml` and the README, but `templates/receive/split/router-deployment.yaml` never calls the helper. In split mode the router is the only workload without `hostUsers`, even with `global.hostUsers=false`. Please add this next to the existing `podSC` include:
> 
> ```yaml
> {{- include "thanos.podHostUsers" (dict "root" . "key" "router" "parent" "receive") | nindent 6 }}
> ```
> 
> **2. Unit tests**
> 
> Please add a few cases under `charts/thanos/tests/`:
> 
>     * no `hostUsers` rendered by default
> 
>     * `global.hostUsers: false` reaches every workload
> 
>     * a component value overrides global (e.g. `global: false`, `query: true`)
> 
>     * the router falls back to `receive.hostUsers` in split mode
> 
> 
> **3. Schema**
> 
> `values.schema.json` doesn't define `hostUsers`, so anything is accepted (e.g. `hostUsers: yes` renders as-is). Please add it as `"type": ["boolean", "null"]` wherever `podSecurityContext` is defined, both global and per component.
> 
> **4. README type**
> 
> The generated table shows `string` / `nil`. Starting the comments with `# -- (bool) Configure hostUsers ...` makes helm-docs show `bool`.
> 
> **Nits (optional)**
> 
>     * In `thanos.podHostUsers`, `$root.Values.nonExistentKey` is used to get a nil value. `{{- $hu := "" -}}` with a `$found` flag, or checking `hasKey` directly, would be easier to read. The indentation of the `else` is also off.
> 
>     * Could you shorten the PR title so it isn't cut off? It becomes the squash commit message. For example: `feat: add hostUsers to pod specs`.
> 
> 
> With these done it's good to go from my side. Thanks again!

Thanks for this very detailed list @hamdikh. As of I'm a newbye in this I need a little time to build the improvements to my PR. I'll mention you here when I'm done. Thanks again, this helps me a lot in my Helm chart journey!

